PR #2, opened 2017-02-16
Pull request to allow the team to inspect and discuss next steps for the initial indexer implementation.

mdellabitta:


mdellabitta:
I think the intention is that this will read records in the index schema, and the writing of them will happen elsewhere. The reason for this is that we want the flexibility of being able to roll up to all the indexed documents and index them all at some point.

PR #8, opened 2017-03-30
Add enrichment functionality, starting with geocoding.

markbreedlove:
Add an EnricherMain class, and add a spatial enrichment, with provisions for more enrichments to be added later.

Known issues:
1. (Major one) Our MAPv4 schema is inadequate for capturing parent geographic features the way we need them for indexing by Elasticsearch. Ingestion 2 (the `Adumbla::CoarseGeocode` enrichment) did not resolve parent features, either. This is something we're continuing to discuss internally. For a given place, we are not inserting parent features like county, state, and country. See the comments in `Spatial.scala`.
2. Given an Avro file with an Avro schema embedded in its header, the EnricherMain does not add it to the Avro files that it writes out. These files end up with fields named `_1` and `_2` instead of `id` and `field`.
3. The Turtle that is output represents blank nodes (of which there are many) differently than Krikri. I don't know if this is a problem.

*I should squash these commits before they're merged!*

### DIFF
--- a/src/main/scala/dpla/ingestion3/EnricherMain.scala
+++ b/src/main/scala/dpla/ingestion3/EnricherMain.scala
@@ -1,0 +1,129 @@
+package dpla.ingestion3
+
+import java.io._
+
+import com.databricks.spark.avro._
+import org.apache.avro.Schema
+import org.apache.log4j.LogManager
+import org.apache.spark.SparkConf
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.types.StructType
+import org.apache.spark.sql.{Row, SparkSession}
+import org.apache.spark.storage.StorageLevel
+import org.eclipse.rdf4j.model.Model
+import org.eclipse.rdf4j.model.util.ModelBuilder
+import org.eclipse.rdf4j.rio._
+import dpla.ingestion3.utils.Utils
+import dpla.ingestion3.enrichments.Spatial
+
+/**
+ * Driver for running an enrichment
+ *
+ * This is a prototype that is hardcoded to run the spatial enrichment,
+ * SpatialEnricher.
+ *
+ * See https://digitalpubliclibraryofamerica.atlassian.net/wiki/display/TECH/Ingestion+3+Storage+Specification#Ingestion3StorageSpecification-S3BucketFolderStructureandContents
+ * with suggestions for how to organize input and output locations. 
+ *
+ * Example invocation:
+ * $ sbt "run-main dpla.ingestion3.EnricherMain local[2] <from> <to>"
+ * ... where <from> is the Avro resource (file / directory) and <to> is the
+ *     destination (directory or s3 folder)
+ *
+ */
+object EnricherMain {
+
+  // Avro schema for our output
+  val schemaStr: String =
+    """
+    {
+      "namespace": "la.dp.avro",
+      "type": "record",
+      "name": "MAPRecord.v1",
+      "doc": "Prototype enriched record",
+      "fields": [
+        {"name": "id", "type": "string"},
+        {"name": "document", "type": "string"}
+      ]
+    }
+    """
+  val logger = LogManager.getLogger(EnricherMain.getClass)
+
+  /*
+   * args:
+   *   master: String -- Spark master descriptor
+   *   inputURI: String -- URI of Avro input
+   *   outputURI: String -- URI of Avro output
+   */
+  def main(args: Array[String]) {
+
+    if (args.length != 3) {
+      logger.error("Arguments should be <master>, <inputURI>, <outputURI>")
+      sys.exit(-1)
+    }
+    val sparkMaster = args(0)
+    val inputURI = args(1)
+    val outputURI = args(2)
+
+    val start_time = System.currentTimeMillis()
+
+    val sparkConf = new SparkConf()
+      .setAppName("Enricher")
+      .setMaster(sparkMaster)
+    val spark = SparkSession.builder().config(sparkConf).getOrCreate()
+    val sc = spark.sparkContext
+
+    val rdd: RDD[Row] = spark.read
+      .format("com.databricks.spark.avro")
+      .option("avroSchema", schemaStr)
+      .load(inputURI)
+      .rdd
+
+    // Process all of the records and wrap them with a dataframe
+    val rows = rdd.map(
+      record => (
+        record.getAs[String]("id"),
+        process(record.getAs[String]("document"))
+      )
+    )
+    // The dataframe is created with fields named "id" and "document" which is
+    // not as nice as actually embedding the schema that we've defined in
+    // schemaStr. I've tried to embed that using
+    // .options("avroSchema", schemaStr) below in the call to dataframe.write(),
+    // but that silently does nothing.
+    // -mb
+    val dataframe = spark.createDataFrame(rows).toDF("id", "document")
+
+    val recordCount = dataframe.count()
+    Utils.deleteRecursively(new File(outputURI))
+    dataframe.write
+             .format("com.databricks.spark.avro")
+             .avro(outputURI)
+    sc.stop()
+
+    val end_time = System.currentTimeMillis()
+    Utils.printResults((end_time - start_time), recordCount)
+
+  }
+
+  def process(rdfString: String): String = {
+    try {
+      val rdfByteStream = new StringReader(rdfString)
+      val in_model = Rio.parse(rdfByteStream, "", RDFFormat.TURTLE)
+      val out_model: Model = Spatial.enrich(in_model)
+      // out_model could be passed to other enrichments in turn
+      val out = new StringWriter()
+      Rio.write(out_model, out, RDFFormat.TURTLE)
+      out.toString
+    } catch {
+      case e: Exception => {
+        val sw = new StringWriter
+        val pw = new PrintWriter(sw)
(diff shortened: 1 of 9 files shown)


markbreedlove:
@mdellabitta: I have some files in S3 that represent the input and output data.

src/main/scala/dpla/ingestion3/EnricherMain.scala
```diff
@@ -0,0 +1,124 @@
...
+             .avro(outputURI)
+    sc.stop()
+
+    val end_time = System.currentTimeMillis()
+    Utils.printResults((end_time - start_time), recordCount)
+
+  }
+
+  def process(rdfString: String): String = {
+    try {
```
mdellabitta: I think in general this is true, with the modification that the mapper code shouldn't itself add a wrapping class like a Try, but this should be knit into the pipeline that calls the mapper to increase flexibility. Which implies that what you did here is the right way to do it for now.

src/main/scala/dpla/ingestion3/EnricherMain.scala
```diff
@@ -0,0 +1,124 @@
...
+      val rdfByteStream = new StringReader(rdfString)
+      val in_model = Rio.parse(rdfByteStream, "", RDFFormat.TURTLE)
+      val out_model: Model = Spatial.enrich(in_model)
+      // out_model could be passed to other enrichments in turn
+      val out = new StringWriter()
+      Rio.write(out_model, out, RDFFormat.TURTLE)
+      out.toString
+    } catch {
+      case e: Exception => {
```
mdellabitta: Ignore for now considering my previous comment about wrappers happening at a higher level.

src/main/scala/dpla/ingestion3/enrichments/Spatial.scala
```diff
@@ -0,0 +1,172 @@
...
+case class Interpretation(feature: Feature)
+
+case class Result(interpretations: List[Interpretation])
+
+object Spatial extends MappingUtils with DefaultVocabularies {
+
+  implicit val formats = DefaultFormats // Formats for json4s
+  val log = Logger.getLogger(getClass.getName)
+
+  def enrich(m: Model): Model = {
```
mdellabitta: I think in general, adding case classes to extract fields is an option, but something you mentioned earlier suggested to me that you thought this was the only way to do it. The JsonExtractionUtils class has an example of doing it using pattern matching, which might end up being more lightweight. You'd probably only implement two case statements, one for extracting the values you expect, and a wildcard one for when something weird happened.

src/main/scala/dpla/ingestion3/enrichments/Spatial.scala
```diff
@@ -0,0 +1,172 @@
...
+
+    val spatialModel = m.filter(null, rdf.`type`, edm.Place)
```
mdellabitta: Codacy is really checking the RDF4J API here, I'd say ignore. 

Although in the long run, maybe having utility functions that have more descriptive names and filter out the nulls might be a good idea for legibility.

src/main/scala/dpla/ingestion3/enrichments/Spatial.scala
```diff
@@ -0,0 +1,172 @@
...
+    if (spatialModel.isEmpty) return m
```
mdellabitta: A way to avoid using return is factor out the meat of this method into a private method and have the if expression essentially be the only thing in this parent method, so:
`if (spatialModel.isEmpty) m
else callPrivateMethod(m)
`

src/main/scala/dpla/ingestion3/enrichments/Spatial.scala
```diff
@@ -0,0 +1,172 @@
...
+
+    for {
+      s <- spatialModel.subjects
+      provLabels = m.filter(s, dpla.providedLabel, null)
```
mdellabitta: Again, Codacy doesn't know we're using a java api here.

src/main/scala/dpla/ingestion3/enrichments/Spatial.scala
```diff
@@ -0,0 +1,172 @@
...
+       * general.
+       *
+       * We need some research to determine how common this is. Ideally we'd
+       * have multiple dc:spatial statements instead of one statement with a
+       * multiline string as the object.
+       *
+       * Is this an issue with Krikri that won't happen with Ingestion 3?
+       */
+
+      try {
```
mdellabitta: Another place where we will eventually want to return a `Try` and make the caller deal with what to do when something fails.

src/main/scala/dpla/ingestion3/enrichments/Spatial.scala
```diff
@@ -0,0 +1,172 @@
...
+
+        // Next do something with parents (requires adding case classes above
+        // for parent features, etc.)
+
+      } catch {
+        // Best way to log a message?
+        case e: java.lang.IndexOutOfBoundsException => {
+          log.error(s"No interpretations for ${pl.getObject.stringValue}")
+        }
+        case e: Throwable => {
```
mdellabitta: `Throwable` is generally too high in the exception hierarchy to catch. I'd probably do `Exception` without knowing more about what else might pop out here.

src/main/scala/dpla/ingestion3/enrichments/Spatial.scala
mdellabitta: Another option is to define the case classes inline where they're being used, so you don't have to go hunting for them, and they don't pollute the namespace. Assuming that this is the only place you'd end up using them.

src/main/scala/dpla/ingestion3/mappers/rdf/Vocabularies.scala
```diff
@@ -39,11 +39,11 @@ case class RDF() extends Vocabulary("rdf", "http://www.w3.org/1999/02/22-rdf-syn
   val HTML = org.eclipse.rdf4j.model.vocabulary.RDF.HTML
 }
 
-case class DPLA() extends Vocabulary("dpla", "http://dp.dpla/about/map/") {
+case class DPLA() extends Vocabulary("dpla", "http://dp.la/about/map/") {
   val Place: IRI = apply("Place")
+  val SourceResource: IRI = apply("SourceResource")
```
mdellabitta: What bonehead made this mistake? :)

src/main/scala/dpla/ingestion3/mappers/rdf/Vocabularies.scala
```diff
@@ -366,7 +366,7 @@ case class ORE()
 }
 
 case class SKOS()
-  extends Vocabulary("skos", "http://www.w3.org/2004/02/skos/core") {
+  extends Vocabulary("skos", "http://www.w3.org/2004/02/skos/core/") {
```
mdellabitta: Ooops!

src/main/scala/dpla/ingestion3/EnricherMain.scala
```diff
@@ -0,0 +1,124 @@
...
+      .rdd
+
+    // Process all of the records and wrap them with a dataframe
+    val rows = rdd.map(
+      record => (
+        record.getAs[String]("id"),
+        process(record.getAs[String]("document"))
+      )
+    )
+    val dataframe = spark.createDataFrame(rows)
```
mdellabitta: Probably a better thing to do here is to call `toDF("id", "document")` on the `RDD`. This is probably the only feedback I have that you should probably implement right away, because right now your resulting DataFrame doesn't have the right column labels and therefore doesn't actually conform to the applied Avro schema.

src/main/scala/dpla/ingestion3/enrichments/Spatial.scala
markbreedlove: I'll change that.

I was responding to a warning from the compiler about how I should change something that I'd written at some point (can't remember exactly what it was) to define `e` as `Throwable`, I think because I may have had it untyped.

mdellabitta: Throwable includes everything, including the `Error` hierarchy, as in `OutOfMemoryError`, etc. The Java API defines those as expressly something that you should not try to catch: https://docs.oracle.com/javase/7/docs/api/

mdellabitta: Correct URL: https://docs.oracle.com/javase/7/docs/api/java/lang/Error.html

src/main/scala/dpla/ingestion3/EnricherMain.scala
markbreedlove: I am getting a compile error calling `.toDF` in multiple ways on the RDD. Gettng:
```
value toDF is not a member of org.apache.spark.rdd.RDD[(String, String)]
```
There are various posts on the web on this issue, but none match my case exactly. (Many of them have to do with case classes being in the wrong place, or using something named SQLCOntext).

Still looking for a solution...

markbreedlove: Got it. I'm going to push a commit that fixes it.

markbreedlove:
Thanks for your help with this, @mdellabitta !

I still have to learn how to handle exceptions the best way in Scala.

mdellabitta:
We should talk about this in more detail during tech staff next week!

PR #9, opened 2017-05-02
Use DataSource API for OAI harvest

AudreyAltman:
This implements the DataSource API for an OAI harvest instead of the RDD API.
I tested this locally using the following params:
```
endpoint: http://oai.artstor.org/oaicatmuseum/OAIHandler
metadataPrefix: oai_dc
verb: ListRecords
```
It addresses [ticket DT-1278](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1278).


AudreyAltman:
Good suggestion @mdellabitta, thanks!  I changed the field name as you suggested rebased.

PR #11, opened 2017-05-04
Use modern elasticsearch-spark in IndexerMain

markbreedlove:
In `IndexerMain`, use the modern, Elasticsearch 5.x-compatible `org.elasticsearch.spark` library instead of the old 0.90-compatible `elasticsearch-hadoop`.

This is a proof-of-concept that is capable of PUTing documents to Elasticsearch, but we have to figure out how to specify an Elasticsearch 5 index mapping in order to run real indexings. See the "FIXME" comment in `IndexerMain.scala`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -10,6 +10,7 @@ lazy val root = (project in file("."))
       "org.scalatest" %% "scalatest" % "3.0.1" % "it,test",
       "org.apache.spark" %% "spark-core" % "2.0.1" exclude("org.scalatest", "scalatest_2.11"),
       "org.apache.spark" %% "spark-sql" % "2.0.1" exclude("org.scalatest", "scalatest_2.11"),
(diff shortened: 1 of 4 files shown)


markbreedlove:
Actually, per https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1282, this should not be merged to `develop`. I guess we just need a review at this point, and we'll keep this on its own branch for a while.

mdellabitta:
RE: your comment, @mb: I think we might need an investigation story to quantify the amount of work it would take to get us to ES 5 before we can decide which way to go on that front.

markbreedlove:
Yes, I'll be writing stories for creating ES 5.x mappings and investigating the move to ES 5

markbreedlove:
Please note that I have just rebased this again on top of `develop`.

PR #12, opened 2017-05-09
get muliple pages from oai harvest

AudreyAltman:
This uses the DataSource interface to harvest multiple pages from an OAI feed.

* The process of collecting HTTP responses from the OAI feed cannot be parallelized due the nature of the OAI feed.  To reduce memory load prior to parallelization, the responses are treated as Strings rather than XML objects.  They are parsed into XML object after parallelization has occurred.
* The `OaiResponseProcessor` has been changed from a class to a static object in order to make it easily serializable.
* At this time, the only two methods of `OaiResponseProcessor` that are being used are `getRecordsAsTuples` and `getResumptionToken`.

This has been tested locally with the following options:
```
--executor-memory "1g" 
--driver-memory "2g"
endpoint: http://oai.artstor.org/oaicatmuseum/OAIHandler
metadataPrefix: oai_dc
verb: ListRecords
```

This addresses [ticket DT-1279](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1279).


mdellabitta:
Great work!

PR #14, opened 2017-05-11
Feature/resource sync initial

moltude:
@AudreyAltman's work on the OAI harvester was branched off here. I think the RS harvester work was working enough for the proof-of-concept it can be merged it but there is some work on this branch


src/main/scala/dpla/ingestion3/utils/Utils.scala
```diff
@@ -61,6 +61,8 @@ object Utils {
   }
 
   /**
+<<<<<<< HEAD
+<<<<<<< HEAD
```
AudreyAltman: Is this leftover from resolving a merge conflict?

AudreyAltman:
I think this is okay to merge.  I have some ideas for how we can keep iterating on this, and I imagine others do as well.  But it seems like a good idea to get this committed to develop.  @moltude do you want to look into the codacy errors and/or squash some of the commits?

AudreyAltman:
Rebase to develop?

PR #15, opened 2017-05-12
Add optional filter arg to IndexderMain

AudreyAltman:
This adds an optional filter argument that can be passed in at the command line.  Only those records that contain the given filter string are indexed.  So for example, if you want to index only items (and not collections), you can pass this argument: `"ingestType": "item"`.

I tested this by filtering over the records as RDDs and as DataFrames.  Using RDDs was about 10% faster.

This has been tested locally.  It addresses ticket [DT-1332](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1332).


AudreyAltman:
This is passing codacity now.

PR #16, opened 2017-05-19
Harvest OAI set

AudreyAltman:
This adds the capability to harvest sets from an OAI feed. Per the scope of this ticket, sets names are passed into the `OaiHarversterMain` from the command line.

A couple points of interest:
* A new class, `OaiResponseBuilder` handles responses to the OAI feed.  It also takes on all of the partitioning logic associated with constructing the requests and handling the responses.  It contains some logic that was previously in `OaiRelation`.
* The process of fetching sets from the OAI feed is parallelized, which greatly improves the speed of harvest.  I harvested all our sets from `artstor` in 18 minutes (128,098 records at 118 records per second).
* I tried re-partitioning the fetched records pages before beginning the process of parsing out XML records (to deal with any potential skew).  This did not help performance, and in fact worsened it a little bit.
* `OaiHarvesterMain`defines a `def main(args: Array[String])` method instead of extending `App` to fix closure issues when serializing.

This has been tested locally.  It addresses [ticket DT-1280](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1280).

### DIFF
--- a/src/main/scala/dpla/ingestion3/harvesters/QueryUrlBuilder.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/QueryUrlBuilder.scala
@@ -22,9 +22,9 @@ class OaiQueryUrlBuilder extends QueryUrlBuilder with Serializable {
     assume(params.get("verb").isDefined)
     assume(params.get("metadataPrefix").isDefined)
 
-    val verb = params.get("verb").get
-    val metadataPrefix = params.get("metadataPrefix").get
-    val url = new URL(params.get("endpoint").get)
+    val verb = params("verb")
+    val metadataPrefix = params("metadataPrefix")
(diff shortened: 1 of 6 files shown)


src/main/scala/dpla/ingestion3/harvesters/QueryUrlBuilder.scala
```diff
@@ -34,8 +34,28 @@ class OaiQueryUrlBuilder extends QueryUrlBuilder with Serializable {
       .setParameter("verb", verb)
 
     params.get("resumptionToken") match {
```
mdellabitta: As much as I <3 pattern matches, I'm wondering if you could use `getOrElse(<default value>)` more in cases like this with Option or Map.

PR #22, opened 2017-06-26
construct OAI URL with port number

AudreyAltman:
This includes the port number when constructing a URL in the OAI harvester.  It has been tested locally with the North Carolina feed.  This addresses [ticket DT-1425](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1425).


moltude:
👍

PR #24, opened 2017-06-27
Feature/string enrichments val syntax

moltude:


mdellabitta:
The Codacy failure for this is annoying because it's fine with the other `type` definition right after it: https://www.codacy.com/app/dpla/ingestion3/pullRequest?bid=4821746&prid=738018&utm_campaign=new_commit&utm_medium=Email&utm_source=Internal&

markbreedlove:
@AudreyAltman @moltude @mdellabitta: Anyone else want to give a review before this is merged?

If we're ready, one of us three committers should squash the changes.

src/main/scala/dpla/ingestion3/enrichments/StringEnrichments.scala
```diff
@@ -0,0 +1,67 @@
...
+  val convertToSentenceCase: SingleStringEnrichment = (value) => {
+    val pattern: Regex = """.*?(\.)""".r
+    val sentences = for( t <- pattern findAllIn value) yield t.trim.capitalize
+    // rejoin the sentences and add back the whitespace that was trimmed off
+    sentences.mkString(" ")
+  }
+
+  val limitCharacters: SingleStringEnrichment = (value) => {
+    ""
+  }
```
AudreyAltman: I'm not sure I understand the use case for `limitCharacters`, but that could be my unfamiliarity with the ingestion process.  In any case, it seems to be the only method without a test.  Does it need one?

mdellabitta: It's basically for gargantuan titles as far as I know. We should probably test it.

mdellabitta: Oh, the feature itself is not implemented. I guess we thought it was so easy, we could get to it later.

AudreyAltman: Maybe add a `TODO` comment to clarify?

AudreyAltman:
The `limitCharacters` implementation looks good!  Thanks @mdellabitta

mdellabitta:
All those years of programming have been leading up to this moment! :)

AudreyAltman:
Rebase to develop?

mdellabitta:
Rebased on develop + squashed.

PR #25, opened 2017-06-29
split oai harvester into three discrete tasks:

AudreyAltman:
This divides the Oai Harvester into three discrete Main tasks:
1. Harvest sets (output = avro dump)
2. Generate set list (input = avro dump of sets; output = comma-separated list of strings printed to command line console)
3. Harvest records (input can include set list generated in step 2)

This accomplished the following objectives:
1. We can now generate a whitelist of all sets to be used in a record harvest (see DT-1426).
2. The interfaces are simpler.  

In future, we may want to build some sort of wrapper around the three harvesting tasks, but for now, this will allow us to complete harvests.

### DIFF
--- a/src/main/scala/dpla/ingestion3/harvesters/QueryUrlBuilder.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/QueryUrlBuilder.scala
@@ -19,20 +19,11 @@ class OaiQueryUrlBuilder extends QueryUrlBuilder with Serializable {
     // Required properties, not sure if this is the right style
     assume(params.get("endpoint").isDefined)
     assume(params.get("verb").isDefined)
-    assume(params.get("metadataPrefix").isDefined)
-
-    /*
-     * The following OAI verbs require a metadataPrefix argument.
-     * If an OAI call contains a verb NOT in this list AND a metadataPrefix,
-     * a badArgument errors is returned.
-     *
-     * @see https://www.openarchives.org/OAI/openarchivesprotocol.html#ProtocolMessages
-     */
-    val verbsWithMetadataPrefix = List("GetRecord", "ListIdentifiers", "ListRecords")
-
-    val verb = params("verb")
-    val metadataPrefix = params("metadataPrefix")
     val url = new URL(params("endpoint"))
+    val verb = params("verb")
+
+    // Optional properties.
+    val metadataPrefix: Option[String] = params.get("metadataPrefix")
     val resumptionToken: Option[String] = params.get("resumptionToken")
     val set: Option[String] = params.get("set")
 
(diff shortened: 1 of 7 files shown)


mdellabitta:
I understand the desire to split these tasks up to make them more composable, but I'm unsure if 3rd party clients of this datasource are going to want to deal with the added complexity of multiple invocations. I'm thinking about this in terms of the mental load of running pyOaiHarvester... there's just one invocation to worry about.

I'm wondering if we could hide the chained invocation of the individual tasks in DefaultSource.scala
 or OaiRelation.scala so that a casual user (not us) would only have that one clean datasource API to think about and not several main classes to invoke to accomplish one task. There's not much reason why you can't create multiple RDDs in there and emit one final RDD with the item data, is there?

Also, I think when we're recording pretty small, simple data, like OAI sets, we could probably get away with CSV or JSON rather than AVRO.

mdellabitta:
I guess I failed to mention that I'm envisioning a scenario where we publish our OAI harvester datasource as a standalone module on https://spark-packages.org/ so that someone who was interested in OAI harvests but not other parts of our tooling could grab our harvester and run it in a notebook or stick it in some other workflow with minimal overhead much as we're using the Crealytics code. A wrapper written in plain old Scala that builds convenience methods might not be legible to all callers in that context.

AudreyAltman:
@mdellabitta I agree that we should be working toward the ability to make a single invocation.  It's difficult now b/c we lack the ability have named and optional parameters from the command line.  So, to cover all our current use cases with a single invocation, we would need 8 distinct parameters, 4 of which _should_ be optional depending on the use case (but instead would have to be entered as some default value like an empty string).  Rather than dealing with that yet, Scott and I thought that having multiple invocations might be a good intermediate step, and that we could prioritize finding a more sophisticated way to handle command-line parameters.  Does that make sense to you?  If not, I can refactor now to make it single-invocation. 

Also, +1 on the simpler file structure for sets.

mdellabitta:
OK, intermediate steps are good. I hadn't thought about how things were coupled to argument parsing. I'll make a story for that.

src/main/scala/dpla/ingestion3/harvesters/QueryUrlBuilder.scala
```diff
@@ -60,12 +51,12 @@ class OaiQueryUrlBuilder extends QueryUrlBuilder with Serializable {
...
     /*
      * Set metadataPrefix param if:
-     *   - the verb allows a metadataPrefix argument AND
+     *   - metadataPrefix is defined
      *   - there is no resumption token (if an OAI call with a resumption token
      *     also contains a metadataPrefix, an OAI badArgument error is returned).
      */
-    if(verbsWithMetadataPrefix.contains(verb) && !resumptionToken.isDefined) {
-      urlParams.setParameter("metadataPrefix", metadataPrefix)
+    if(metadataPrefix.isDefined && !resumptionToken.isDefined) {
```
mdellabitta: A way to get Codacy to shut up about this:

`if (!resumptionToken.isDefined) 
    metadataPrefix.foreach(prefix => urlParams.setParameter("metadataPrefix", prefix))
`
Not sure it's worth it, though. In any case, this is an interesting blog: https://blog.protegra.com/2014/01/28/exploring-scala-options/

AudreyAltman:
@mdellabitta I made the codacy fix per your suggestion -- thanks!

AudreyAltman:
Oh shoot, I didn't fix the codacy problem.  One moment...

PR #26, opened 2017-06-29
Fix to HTTP redirection issue. Subbed in HttpComponent's fluent reque…

mdellabitta:
Fix to HTTP redirection issue that appeared as an XML parsing issue. Subbed in HttpComponent's  fluent client in place of IOUtils for fetching OAI results since it follows redirects. Should consider a better HTTP library still because HttpClient can get pretty enterprisey.


markbreedlove:
I see what you mean about the "enterpriseyness," but I'm OK with this for now.

moltude:
👍 Are we doing in-browser rebasing or @mdellabitta would you prefer to do that locally?

PR #27, opened 2017-06-30
Feature/dt 1424 add blacklist filter

moltude:
Accepts a blacklist as a comma separated string and removes blacklisted values from the set list
Removed scrapwood.sc from the project


AudreyAltman:
@moltude technically I don't believe you can't have two optional inputs from the command line, but since we're going to change it anyway, I think it's okay for now.

PR #28, opened 2017-06-30
Mapping using model objects and an RDF transformation

mdellabitta:
The refactor to generify the translation of our domain objects to RDF and make the only provider-specific part of mapping the extraction from the provider document to our schema.


markbreedlove:
Hi, @mdellabitta, I'm prepared to start a review, but first was wondering if you wanted to comment on the stuff that Codacy's flagging.

mdellabitta:
A good chunk of it is carryover from stuff that already existed. 

The imports ones are based on having individual lines for each import rather than grouping them together or having a wildcard. That was a preference of yours, @markbreedlove, but I think since the impact of plain imports at the top of a file is different in Scala than it is in Python, we should probably decide that as a group. IntelliJ with out of the box defaults will automatically combine these with Code > Optimize Imports and I assume it would make Codacy happy.

This particular "put imports at the top of the file" is a good one. I think there's a case for not doing so when using typeclasses and other higher-order constructs, but we haven't got that fancy yet, and probably won't.

"object/type is a reserved keyword", the alternative is to come up with some name that looks close enough that you can recognize what's being said, like "clazz" for "class." I was trying to match the MAP specification, and so used backticks to use the literal name. Open to feedback.

"Field name 'DCMI_TYPES' does not match the regular expression" is also valid. It looks like I should have done this: https://stackoverflow.com/questions/9745488/naming-convention-for-scala-constants

mdellabitta:
I forgot the "longer than 50 lines" one. I think this is less of an issue when writing seemingly declarative, plain code, but again, open to feedback.

mdellabitta:
This last push should have made most of the Codacy warnings go away, except for the one about reusing `object` and `type` as variable names. Still could go either way on that one, but it would be easy to change at any point.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -0,0 +1,145 @@
...
+                               publisher: ZeroToMany[EdmAgent] = Seq(),
+                               relation: ZeroToMany[LiteralOrUri] = Seq(),
+                               replacedBy: ZeroToMany[String] = Seq(),
+                               replaces: ZeroToMany[String] = Seq(),
+                               rights: AtLeastOne[String] = Seq(),
+                               rightsHolder: ZeroToMany[EdmAgent] = Seq(),
+                               subject: ZeroToMany[SkosConcept] = Seq(),
+                               temporal: ZeroToMany[EdmTimeSpan] = Seq(),
+                               title: AtLeastOne[String] = Seq(),
+                               `type`: ZeroToMany[String] = Seq() //should be URIs?
```
markbreedlove: It's probably OK to call this dcmiType or dctermsType to avoid having to quote it.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -0,0 +1,145 @@
...
+  *
+  */
+
+case class OreAggregation(
+                           uri: ExactlyOne[URI], //uri of the record on our site
+                           dataProvider: ExactlyOne[EdmAgent],
+                           originalRecord: ExactlyOne[String], //map v4 specifies this as a ref, but that's LDP maybe?
+                           hasView: ZeroToMany[EdmWebResource] = Seq(),
+                           intermediateProvider: ZeroToOne[EdmAgent] = None,
+                           `object`: ZeroToOne[EdmWebResource] = None, //full item page
```
markbreedlove: Same as above, should this be named edmObject?

(What's worse, having backticks, or having two variable names that don't match the field names in the API where most of the other variable names do match?)

mdellabitta: I hate MySQL, so the backticks are a little unnerving.

moltude: IMO I'd be more concerned with having the fields match the API, this is the kind of thing that burns us with ingestion1 until it is burned into your brain. (e.g. object/preview)

mdellabitta: I'm now wondering if I just prefix these variables with an underscore, if that'll make the warning go away, and be "close enough."

src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala
```diff
@@ -0,0 +1,84 @@
...
+import dpla.ingestion3.mappers.json.JsonExtractionUtils
+import dpla.ingestion3.model.{DplaMapData, DplaSourceResource, EdmWebResource, OreAggregation, _}
+import org.json4s._
+import org.json4s.jackson.JsonMethods._
+
+class CdlExtractor extends Extractor with JsonExtractionUtils {
+
+  def build(rawData: String): DplaMapData = {
+
+    implicit val json: JValue = parse(rawData)
```
AudreyAltman: Why is this an implicit val, rather than just a val?

mdellabitta: So I don't have to pass it as an argument explicitly. Makes the following code cleaner looking.

src/main/scala/dpla/ingestion3/mappers/providers/PaExtractor.scala
```diff
@@ -0,0 +1,61 @@
...
+      ),
+
+      EdmWebResource(
+        uri = itemUrl,
+        fileFormat = extractStrings("dc:format")
+      ),
+
+      OreAggregation(
+        uri = new URI("http://example.com"), //todo our url
+        //below will throw if not enough contributors
```
AudreyAltman: Is this the desired behavior?

mdellabitta: Yeah, dataProvider is required.

src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala
AudreyAltman: I see, :+1:

PR #29, opened 2017-07-05
Add named command line arguments to OAI harvester

moltude:
Uses Scallop (https://github.com/scallop/scallop/wiki) to implement named command line arguments for the OAI harvester. 

Arguments to be passed into the OAI-PMH harvester
  * Output directory: String
  * OAI URL: String
  * Metadata Prefix: String oai_dc oai_qdc, mods, MODS21 etc.
  * Provider: Provider name (we need to standardize this).
  * Sets (Optional): Comma-separated String of sets to harvest from.
  * Blacklist (Optional): Comma-separated String of sets to exclude from harvest
  * sparkMaster (Optional): If omitted it defaults to local[*]

### DIFF
--- a/src/main/scala/dpla/ingestion3/OaiRecordHarvesterMain.scala
+++ b/src/main/scala/dpla/ingestion3/OaiRecordHarvesterMain.scala
@@ -4,6 +4,7 @@ import java.io.File
 
 import dpla.ingestion3.utils.Utils
 import com.databricks.spark.avro._
+import dpla.ingestion3.confs.OaiHarvesterConf
 import org.apache.log4j.LogManager
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.SparkSession
@@ -42,43 +43,33 @@ object OaiRecordHarvesterMain {
   val logger = LogManager.getLogger(OaiRecordHarvesterMain.getClass)
 
   def main(args: Array[String]): Unit = {
-
-    validateArgs(args)
-    println(schemaStr)
-
-    val outputFile = args(0)
-    val endpoint = args(1)
-    val metadataPrefix = args(2)
-    val provider = args(3)
-
-    // This is an Option (as opposed to a String) b/c the param args(4) is optional.
-    val sets: Option[String] = if (args.isDefinedAt(4)) Some(args(4)) else None
-
-    // TODO rewrite this will better named parameters
-    val blacklistSets: Option[String] = if (args.isDefinedAt(5)) Some(args(5)) else None
-
+    val oaiConf = new OaiHarvesterConf(args.toSeq)
     val verb = "ListRecords"
 
-    Utils.deleteRecursively(new File(outputFile))
+    Utils.deleteRecursively(new File(oaiConf.outputDir()))
 
     // Initiate spark session.
     val sparkConf = new SparkConf().setAppName("Oai Record Harvest")
+    // sparkMaster has a default value of local[*] if not provided.
+    // TODO: will this default value work with EMR?
+    sparkConf.setMaster(oaiConf.sparkMaster())
+
     val spark = SparkSession.builder().config(sparkConf).getOrCreate()
     val sc = spark.sparkContext
 
     val start = System.currentTimeMillis()
 
-    val baseOptions = Map("metadataPrefix" -> metadataPrefix, "verb" -> verb)
-    val readerOptions = getReaderOptions(baseOptions, sets, blacklistSets)
+    val baseOptions = Map("metadataPrefix" -> oaiConf.prefix(), "verb" -> verb)
+    val readerOptions = getReaderOptions(baseOptions, oaiConf.sets.toOption, oaiConf.blacklist.toOption)
 
     val results = spark.read
       .format("dpla.ingestion3.harvesters.oai")
       .options(readerOptions)
-      .load(endpoint)
+      .load(oaiConf.endpoint())
 
     results.persist(StorageLevel.DISK_ONLY)
 
-    val dataframe = results.withColumn("provider", lit(provider))
+    val dataframe = results.withColumn("provider", lit(oaiConf.provider()))
       .withColumn("mimetype", lit("application_xml"))
 
     val recordsHarvestedCount = dataframe.count()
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/OaiRecordHarvesterMain.scala
```diff
@@ -42,27 +43,27 @@ object OaiRecordHarvesterMain {
   val logger = LogManager.getLogger(OaiRecordHarvesterMain.getClass)
 
   def main(args: Array[String]): Unit = {
+    val oaiConf = new OaiHarvesterConf(args.toSeq)
 
-    validateArgs(args)
-    println(schemaStr)
+    // These are the minimum required parameters for an OAI harvest
+    val outputFile = oaiConf.outputDir.apply()
```
mdellabitta: For these apply() calls, the shorthand is just (). so you could change:

`oaiConf.outputDir.apply()` 

to 

`oaiConf.outputDir`

I'm thinking that since the oaiConf fields are named pretty obviously, you could avoid creating these variables here and just sub in the call to oaiConfDir.someField().

src/main/scala/dpla/ingestion3/OaiRecordHarvesterMain.scala
```diff
@@ -42,27 +43,27 @@ object OaiRecordHarvesterMain {
...
+    val blacklistSets: Option[String] = oaiConf.blacklist.toOption
+    val sparkMaster: String = oaiConf.sparkMaster.apply()
 
     val verb = "ListRecords"
 
     Utils.deleteRecursively(new File(outputFile))
 
     // Initiate spark session.
     val sparkConf = new SparkConf().setAppName("Oai Record Harvest")
+    sparkConf.setMaster(sparkMaster)
```
mdellabitta: Looks like if this stays in, we're running this stuff through sbt rather than spark-submit. I like that idea for now, but not sure if it's workable when using EMR or not.

src/main/scala/dpla/ingestion3/OaiRecordHarvesterMain.scala
```diff
@@ -121,6 +118,11 @@ object OaiRecordHarvesterMain {
       case (Some(sets), Some(blacklist)) => {
         baseOptions + ("sets" -> (sets.split(",") filter (!blacklist.split(",").contains(_))).mkString(","))
       }
+      // No set list passed in but a blacklist was passed in
+      // TODO: Should this error out or should it try to call ListSets to generate a set list?
```
mdellabitta: I would think that blacklists always imply a ListSets call, no?

src/main/scala/dpla/ingestion3/OaiRecordHarvesterMain.scala
moltude: 👍

src/main/scala/dpla/ingestion3/OaiRecordHarvesterMain.scala
moltude: I agree that you can't have a blacklist without a set list. So the correct response to this condition should be to raise an error.

mdellabitta: I'm confused. If you have a set list locally, you wouldn't need a blacklist because you could just remove the items manually.

AudreyAltman: FWIW I'm working on a refactor that will sort this all out -- it will make all four of these actions available as you laid them out through a single main method.

moltude: I guess the assumption here is that the notion of "sets" should have been resolved before invoking the OaiHarvestMain. To me, it feels like this might be informed by how we set up the refactor of Audrey's work (splitting this into three mains) to include the generation of the sets before invoking the record harvest.  Options 1 - we plan for ListSets to invoked before harvesting records which would simplify the record harvest because sets are passed in as comma separated string if they are provided as a manual "whitelist" in the profile or command line or by calls to ListSets. 

Or we only make a call to ListSets if they gave us a blacklist and no sets.  IMO the first seems option seems more straightforward. 

Not sure how well I've explained my thinking here..

src/main/scala/dpla/ingestion3/OaiRecordHarvesterMain.scala
AudreyAltman: What's the benefit of running this through sbt rather than spark-submit?

src/main/scala/dpla/ingestion3/OaiRecordHarvesterMain.scala
mdellabitta: I'm basing my comments on the "single invocation" idea that we talked about last week and that Audrey seems to be working on. Probably should put these both in the same PR?

AudreyAltman: I would be happy to work my refactor into this PR, but the ticket to create a single-invocation OAI harvest isn't in this sprint: https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1439  We could merge this PR as an "incremental" step, or we could leave it open until I can start work on DT-1439.  Either way is okay with me.

src/main/scala/dpla/ingestion3/confs/OaiHarvesterConf.scala
```diff
@@ -0,0 +1,54 @@
...
+    noshort = true,
+    // TODO is there a better validation of the URL rather than nonEmpty?
+    validate = (_.nonEmpty))
+  val prefix: ScallopOption[String] = opt[String]("prefix",
+    required = true,
+    noshort = true,
+    validate = (_.nonEmpty))
+  val provider: ScallopOption[String] = opt[String]("provider",
+    required = true,
+    noshort = true)
```
AudreyAltman: Should `provider` have a `nonEmpty` validation as well?

moltude:
@AudreyAltman added missing validation to `provider` and rebased.

AudreyAltman:
Rebase to develop?

moltude:
@AudreyAltman ready for merge!

PR #32, opened 2017-07-11
Remove DPLA id generation from OAI, API and RS harvesters.

moltude:
This ID generation will happen during mapping so the harvesters can be utilized outside DPLA.


AudreyAltman:
Looks like you need to edit the spec to reflect the change.

src/main/scala/dpla/ingestion3/harvesters/Harvester.scala
```diff
@@ -1,38 +1,21 @@
...
+    * Accepts a record from an OAI feed an returns the OAI identifier
     *
     * @param record String
     *               The original record from the OAI feed
     * @return The local identifier
     */
-  def getLocalId(record: Node): String = {
+  def getOaiIdentifier(record: Node): String = {
     (record \\ "header" \\ "identifier").text
   }
```
AudreyAltman: Could the above method be moved to `OaiResponseProcessor`?  There is [similar logic](https://github.com/dpla/ingestion3/blob/964d4db7c6a635b371cf127cb0d21ea6db777e38/src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala#L54) in `OaiResponseProcessor` already for getting a set ID, so it would make sense that the logic for getting a record ID follow the same pattern.  Also, I believe that this way we wouldn't need the `Harvester` class at all.

PR #33, opened 2017-07-12
Harvest OAI with single invocation

AudreyAltman:
This enables OAI harvests through a single, command-line invocation.  It satisfies the following requirements:

1. Harvest all sets.
2. Harvest all records.
3. Harvest records from all sets (this is different from (2) b/c not all records are associated with sets, and some records are associated with more than one set)
4. Harvest records from a given setlist (of one or more sets).
5. Harvest records from all sets except those in a given blacklist.

When harvesting records, the resulting dataframe always includes new columns for `set_id` and `set_document`.  If a record has no associated set, the values in these columns are `null`.  This PR introduces case classes for `record` and `set` to make it easier to compile data for the final dataframe.

This has been tested locally.  It addresses [DT-1439](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1439).

One specific question I have is, in which cases (if any) should the app throw `OaiHarvesterException`s?


src/main/scala/dpla/ingestion3/OaiHarvesterMain.scala
```diff
@@ -80,38 +105,13 @@ object OaiRecordHarvesterMain {
       .option("avroSchema", schemaStr)
       .avro(oaiConf.outputDir())
 
+    println(dataframe.show) // For testing only, can delete later.
```
AudreyAltman: I'll just delete it, I don't think it's worth logging.  But good suggestion for future debugging!

AudreyAltman:
@moltude I fixed the schema namespace, deleted the un-necessary `println` statement, and fixed the big I accidentally introduced with my last commit.  If all tests pass, it should be good to go.

moltude:
looks good now! 👍

PR #37, opened 2017-07-19
First crack at fatal error handling

moltude:
* Updates configuration of log4j to send error messages to console
* Adds `endpoint` param to `readerOptions` so it can be validated in DefaultSource
* Wraps the invocation of the harvest in a try/catch block
* Adds validation to DefaultSource for the `verb` and `endpoint` params. For `verb` the parameter must be provided and for right now can only be either `ListRecords` or `ListSets`. Other verbs may be supported at a later date. For `endpoint` it checks that a URL object can be created from the string and the url is reachable. 
* Adds a `validateUrl` method to Utils that performs the endpoint validation described above.

### DIFF
--- a/src/main/scala/dpla/ingestion3/OaiHarvesterMain.scala
+++ b/src/main/scala/dpla/ingestion3/OaiHarvesterMain.scala
@@ -7,10 +7,11 @@ import com.databricks.spark.avro._
 import dpla.ingestion3.confs.OaiHarvesterConf
 import org.apache.log4j.LogManager
 import org.apache.spark.SparkConf
-import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.apache.spark.sql.functions._
 import org.apache.spark.storage.StorageLevel
 
+import scala.util.{Failure, Success, Try}
 
 /**
   * Entry point for running an OAI harvest.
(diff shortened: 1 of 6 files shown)


AudreyAltman:
I think `VocabEnforcer` and `VocabEnforcerTest` are in here on accident?

src/main/scala/dpla/ingestion3/harvesters/oai/DefaultSource.scala
```diff
@@ -31,17 +36,44 @@ class DefaultSource extends RelationProvider {
     new OaiRelation(endpoint, verb, metadataPrefix, harvestAllSets, setlist, blacklist)(sqlContext)
   }
 
+  def getMetadataPrefix(parameters: Map[String, String]): String = {
+    (parameters.get("metadataPrefix"), parameters.get("verb")) match {
+      case (Some(prefix), Some(verb)) => prefix
```
AudreyAltman: I'm not sure we need this validation.  The `getVerb` validation you wrote before nicely checks for the presence of a supported verb.  In absence of a verb, the harvest will fail anyway.

src/main/scala/dpla/ingestion3/OaiHarvesterMain.scala
```diff
@@ -73,56 +75,57 @@ object OaiHarvesterMain {
...
-    readerOptions("verb") match {
-      // Write records to avro.
-      // This task may require a large amount of driver memory.
-      case "ListRecords" => {
-        println(recordSchemaStr)
+    def runHarvest(): Try[DataFrame] = {
+      Try(spark.read
+        .format("dpla.ingestion3.harvesters.oai")
+        .options(readerOptions)
+        .load(oaiConf.endpoint()))
```
moltude: 👍 thanks!

src/main/scala/dpla/ingestion3/harvesters/oai/DefaultSource.scala
moltude: makes good sense to me! Thanks

moltude:
@AudreyAltman I've cleaned it up following your suggestions and somehow (I have no idea how) removed the Vocab* enrichment files from this PR.

src/main/scala/dpla/ingestion3/OaiHarvesterMain.scala
```diff
@@ -73,56 +74,57 @@ object OaiHarvesterMain {
...
-      // Write sets to csv.
-      case "ListSets" => {
-        println(setSchemaStr)
-
-        dataframe.coalesce(1).write
-          .format("com.databricks.spark.csv")
-          .option("header", true)
-          .csv(oaiConf.outputDir())
+    runHarvest() match {
+      case Success(results) => {
```
mdellabitta: Trivia: You don't need braces around multiline case statements.

PR #38, opened 2017-07-19
Feature/lexvo enrichment

moltude:
Initial stab at a lexvo enrichment. May require refinement as the language in DPLA MAP is SKOS not String.

### DIFF
--- a/src/main/scala/dpla/ingestion3/enrichments/VocabEnforcer.scala
+++ b/src/main/scala/dpla/ingestion3/enrichments/VocabEnforcer.scala
@@ -1,7 +1,8 @@
 package dpla.ingestion3.enrichments
 
-import dpla.ingestion3.mappers.rdf.DCMIType
 import org.eclipse.rdf4j.model.IRI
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/enrichments/VocabEnforcer.scala
```diff
@@ -98,3 +100,26 @@ object DcmiTypeMapper extends VocabEnforcer[String] {
...
+  * Reads in ISO-693-3 data from text file and generates a
+  * map of abbreviations >> normalized names.
+  *
+  */
+object LexvoEnforcer extends VocabEnforcer[String] {
+  val iso_639_3 = ISO_639_3()
+
+  val lexvoStrings = {
+    // TODO make this a config property
+    val bufferedSource = Source.fromFile("./data/iso-639-3/iso-639-3.tab")
```
mdellabitta: We might need to figure out another way to get at this data since that file will end up in a JAR at some point. It's something like `getClass().getResource("somepath")`, but we can figure it out later.

PR #39, opened 2017-07-20
always get set data when harvesting records

AudreyAltman:
Before, for an OAI harvest, set data was collected for each record ONLY in the following circumstances:

1. A setlist is given
2. A set blacklist is given
3. `harvestAllSets` is set to `true`

This PR gets set data for records by default.

This has been tested locally with a hub in which all records belong to at least one set (Illinois), and a hub in which sets are not implemented (Kentucky).  This has _not_ been tested for a hub in which only _some_ records belong to sets, because such a hub does not currently exist.

This addresses [ticket DT-1472](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1472).


moltude:
This is pretty small PR so I'm going to merge it w/o a second review. @mdellabitta okay?

PR #40, opened 2017-07-20
Inital work on enrichment driver

moltude:
Required changes to lexvo, date and type enrichments may be included in this branch.
Added tests for lexvo enrichment
Removed ISO-639 vocab case class
Moved iso-data file to `resources` directory


src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
```diff
@@ -5,12 +5,15 @@ import java.text.SimpleDateFormat
...
 import scala.util.{Failure, Success, Try}
 import ParseDateEnrichment._
+import dpla.ingestion3.model.EdmTimeSpan
 
 class ParseDateEnrichment {
 
   //TODO ranges
-  def parse(dateString: String, allowInterval: Boolean = false): Option[String] = {
+  def parse(edtfOriginalDate: EdmTimeSpan, allowInterval: Boolean = false): EdmTimeSpan = {
+    val dateString = edtfOriginalDate.originalSourceDate.get
```
AudreyAltman: Is `originalSourceDate` an Option?  If so, does this need to be a `getOrElse` or something in case the value is None?

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -137,7 +140,19 @@ case class EdmPlace(
...
 case class EdmTimeSpan(
-                        originalSourceDate: ZeroToMany[String] = Seq(),
+                        /* Why is this 0.n? I would have assumed ExactlyOne or ZeroToOne
+                            since it wouldn't make sense to have multiple original ranges
+                            per EdmTimeSpan but rather one edmTimeSpan per originalSourceDate?
+
+                            Esp. since in date in sourceResource is also ZeroToMany...
+
+                            I'm changing the type for the enrichment driver and need to follow-up
+                            w/team
```
AudreyAltman: I don't know enough to contribute to this discussion at this point, so you should probably follow up with Mark or Michael.

src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
```diff
@@ -35,11 +38,20 @@ class ParseDateEnrichment {
     //this iterates over the list, executes the next function, and returns it if there's a result.
     //otherwise, it continues until the end and returns None.
     //I wasn't able to find a better way to emulate ||= from Ruby. There's probably a better way.
```
AudreyAltman: Off the top of my head, I can't come up with a better way.  This way makes sense to me.

src/main/scala/dpla/ingestion3/enrichments/EnrichmentDriver.scala
```diff
@@ -0,0 +1,60 @@
...
+    * Applies a set of common enrichments that need to be run for all providers
+    *   * Spatial
+    *   * Language
+    *   * Type
+    *   * Date
+    *
+    * @param record The mapped record
+    * @return An enriched record
+    */
+  def enrich(record: DplaMapData): DplaMapData = {
```
mdellabitta: what @AudreyAltman  said... you should only have to specify the fields you're going to change. You might need to make a new DplaMapData, though, but that only has 3 fields.

src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
mdellabitta: Yeah, Codacy's going to yell about this for this reason.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
mdellabitta: I was following the MAP v4 spec, which specifies it that way. I believe that @guegueng was hazy about that too. I'm for being pragmatic like you just were, though, @moltude.

src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
mdellabitta: This was me. I still don't feel great about this!

mdellabitta:
I think if you could do the `copy` thing and not call `get()` on the `Option` things that @AudreyAltman mentioned, this is good to merge!

moltude:
Implemented those changes and added a test

AudreyAltman:
@moltude want to rebase and then I'll merge?

PR #41, opened 2017-07-21
DO NOT MERGE: Update avro schemas

moltude:
Moved Avro schema files into separate files
Updated schemas and namespace to match specs

### DIFF
--- a/src/main/resources/avro/IndexRecord_MAP3.avsc
+++ b/src/main/resources/avro/IndexRecord_MAP3.avsc
@@ -1,5 +1,5 @@
 {
-  "namespace": "la.dp.avro.MAP.3_1",
+  "namespace": "dpla.avro.v1.MAP3_1",
   "type": "record",
   "name": "IndexRecord.v1",
   "doc": "",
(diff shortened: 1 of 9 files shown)


moltude:
Bumping for follow-up at sprint review/planning. @markbreedlove Just need to confirm whether there was any additional work required before opening this up for review.

markbreedlove:
I've just realized that EdmPlace should no longer be named as such. We should name this DplaPlace, because I changed it to be DplaPlace from the MAP3.1 schema. `edm:Place` is a specific thing that this is not.

mdellabitta:
Good to know!

markbreedlove:
Actually, it's OK in this PR ... I made this comment in the wrong PR.

markbreedlove:
@moltude: Thanks for your commit to assign the Avro schemas from their files, instead of having them hardcoded.  I will close this PR and squash the branch so that @AudreyAltman can work with it for https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1478 .

I don't understand the reason for the `println` statements in OaiHarvesterMain -- are they for debugging? -- but they are no issue within the context of this change, which is simply to change the source of the schema string.

PR #42, opened 2017-07-21
Dt 1466 oai unit tests

AudreyAltman:
This adds unit tests for three classes in the OAI harvester.  We can add tests for other classes later -
 this was a timeboxed ticket.  This also adds ScalaMock.  There are other mock testing frameworks available if we decide we don't like ScalaMock - I only needed it here for a very simple use case.

This addresses [DT-1471](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1471).

### DIFF
--- a/src/test/scala/dpla/ingestion3/harvesters/oai/DefaultSourceTest.scala
+++ b/src/test/scala/dpla/ingestion3/harvesters/oai/DefaultSourceTest.scala
@@ -1,0 +1,74 @@
+package dpla.ingestion3.harvesters.oai
+
+import org.scalatest._
+
+/**
+  * Tests for DefaultSourceTest
+  */
+class DefaultSourceTest extends FlatSpec {
+  val source = new DefaultSource
+  val endpoint = "http://repox.example.edu:8080/repox/OAIHandler"
+  val verb = "ListRecords"
+  val list = "manuscripts, newspapers, maps"
+  val prefix = "mods"
+  val token = "1234"
+
+  "getEndpoint" should "throw Exception if endpoint missing" in {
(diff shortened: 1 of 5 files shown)


src/test/scala/dpla/ingestion3/harvesters/oai/DefaultSourceTest.scala
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  "getSetlist" should "return None if setlist missing" in {
+    val params = Map("foo" -> "bar")
+    assert(source.getSetlist(params) === None)
+  }
+  it should "return set list as Some[Array[String]]" in {
+    val params = Map("setlist" -> list)
+    // TODO: Is there a better way to match Some[Array[String]]?
+    assert(source.getSetlist(params).get.toSeq === Array("manuscripts", "newspapers", "maps").toSeq)
```
moltude: This doesn't need the additional transformation `.toSeq` just calling get() is enough for a valid comparison but I am confused why you can't compare successfully compare two options. The values are identical but `getSetList(params) === Some(Array(...))` fails

mdellabitta: I don't actually know why that wouldn't work, either!

mdellabitta:
This looks great to me!

Reminds me that we should get Codacy set up to calculate and display test coverage...

moltude:
@AudreyAltman do you want to squash these up before we merge?

AudreyAltman:
Squashed, and coday error fixed.  Ready to merge!

PR #43, opened 2017-07-24
Standalone implementation of NARA file harvesting. 

mdellabitta:
I built a tar DataSource here. Unfortunately, it requires as much memory per tar file as the uncompressed parts of the tar. I've committed it as part of this pull request in case we think it would be useful to keep around.

The working NARA harvest is a standalone, non-Spark command line invocation that converts their tar files of big XML files to Avros of individual item records.

### DIFF
--- a/src/main/scala/dpla/ingestion3/harvesters/file/NaraFileHarvestMain.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/file/NaraFileHarvestMain.scala
@@ -1,0 +1,217 @@
+package dpla.ingestion3.harvesters.file
+
+import java.io.{File, FileInputStream}
+import java.util.zip.GZIPInputStream
+
+import dpla.ingestion3.utils.{FlatFileIO, Utils}
+import org.apache.avro.Schema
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/harvesters/file/NaraFileHarvestMain.scala
```diff
@@ -0,0 +1,224 @@
...
+import org.apache.tools.tar.TarInputStream
+import org.rogach.scallop.{ScallopConf, ScallopOption}
+
+import scala.util.{Failure, Success, Try}
+import scala.xml.{MinimizeMode, Node, Utility, XML}
+
+
+object NaraFileHarvestMain {
+
+  val schemaStr: String =
```
moltude: @mdellabitta after PR #46 is merged in this should pull from those files, correct?

mdellabitta: AKA yes. :)

mdellabitta:
This is rebased on develop and contains bindings for the originalrecord schema.

PR #44, opened 2017-07-25
ID minting 

moltude:
Implements DPLA ID mapping in provider extractors.

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala
@@ -7,12 +7,10 @@ import dpla.ingestion3.model.{DplaMapData, DplaSourceResource, EdmWebResource, O
 import org.json4s._
 import org.json4s.jackson.JsonMethods._
 
-class CdlExtractor extends Extractor with JsonExtractionUtils {
+class CdlExtractor(rawData: String) extends Extractor with JsonExtractionUtils {
+  implicit val json: JValue = parse(rawData)
 
-  def build(rawData: String): DplaMapData = {
-
-    implicit val json: JValue = parse(rawData)
-
+  def build: DplaMapData = {
     DplaMapData(
       DplaSourceResource(
         alternateTitle = extractStrings("alternative_title_ss"),
@@ -42,7 +40,7 @@ class CdlExtractor extends Extractor with JsonExtractionUtils {
         uri = providerUri(json)
       ),
       OreAggregation(
-        uri = new URI("http://example.com"), //TODO: our url
+        uri = mintDplaItemUri(),
         dataProvider = nameOnlyAgent(provider(json)),
         originalRecord = rawData,
         preview = thumbnail(json),
@@ -81,4 +79,6 @@ class CdlExtractor extends Extractor with JsonExtractionUtils {
     name = Some("California Digital Library"),
     uri = Some(new URI("http://dp.la/api/contributor/cdl"))
   )
+  // Base CDL ID should use provider prefix (e.g. cdl--104014afdkfg2a)
+  override def getProviderBaseId(): Option[String] = Option(List("cdl", extractString("id")(json)).mkString("--"))
 }
--- a/src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
@@ -1,12 +1,40 @@
 package dpla.ingestion3.mappers.providers
 
+import java.net.URI
+
 import dpla.ingestion3.model.{DplaMapData, EdmAgent}
+import dpla.ingestion3.utils.Utils
+import org.apache.pig.data.utils.MethodHelper.NotImplemented
 
 /**
   * Interface that all provider extractors implement.
   */
 
 trait Extractor {
-  def build(rawData: String): DplaMapData
+  // Base item uri
+  private val baseItemUri = "http://dp.la/api/items/"
+
+  def build(): DplaMapData
   def agent: EdmAgent
+  /**
+    * Build the base ID to be hashed. Implemented per provider
+    *
+    * @return String
+    * @throws NotImplemented
+    */
+  def getProviderBaseId(): Option[String]
+
+  /**
+    * Hashes the base ID
+    *
+    * @return MD5 hash of the base ID
+    */
+  protected def mintDplaId(): String = Utils.generateMd5(getProviderBaseId())
+
+  /**
+    * Builds the item URI
+    *
+    * @return URI
+    */
+  protected def mintDplaItemUri(): URI = new URI(s"${baseItemUri}${mintDplaId()}")
 }
--- a/src/main/scala/dpla/ingestion3/model/DplaMapData.scala
+++ b/src/main/scala/dpla/ingestion3/model/DplaMapData.scala
@@ -3,7 +3,6 @@ package dpla.ingestion3.model
 import java.net.URI
 
 import dpla.ingestion3.model.DplaMapData._
-import org.eclipse.rdf4j.model.IRI
 
 /**
   * Contains type definitions that express cardinality of fields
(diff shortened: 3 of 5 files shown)


moltude:
Created PR for discussion only. Need to rebase and squash before merge.

src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala
```diff
@@ -42,12 +43,13 @@ class CdlExtractor extends Extractor with JsonExtractionUtils {
         uri = providerUri(json)
       ),
       OreAggregation(
-        uri = new URI("http://example.com"), //TODO: our url
+        uri = new URI(List(DplaMapData.baseItemUri, mintItemId(rawData)).mkString("")),
```
mdellabitta: I think this deserves it's own function, or the mkString call should be moved into mintItemId.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -31,7 +32,9 @@ object DplaMapData {
 case class DplaMapData(
                          sourceResource: DplaSourceResource,
                          edmWebResource: EdmWebResource,
-                         oreAggregation: OreAggregation
+                         oreAggregation: OreAggregation,
+                         // Add id property at the root level (following ES schema)
+                         id: ExactlyOne[String]
```
mdellabitta: Aren't ids URIs?

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
```diff
@@ -9,4 +10,14 @@ import dpla.ingestion3.model.{DplaMapData, EdmAgent}
...
+  /**
+    * Generates an MD5 checksum from the given providerId
+    * Implemented per provider
+    *
+    * DigestUtils.md5Hex(id)
+    *
+    * @return MD5 hash value
+    * @throws NotImplemented
+    */
+  def mintItemId(rawData: Any): String
```
moltude: Is there a better way to generically type the parameter? We want to send a NodeSeq or a JValue (possibly something else?). I've tried using `type T` and `Either[JValue, NodeSeq]` w/o success. 

The original issue I ran into was re-parsing the rawData twice when it was being passed in as a String (like it is done in `build()`) in each of the implementations and that looks pretty ugly. I also didn't want to repeat how the base ID was being selected. That should only exist in the provider implementation once.

src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala
```diff
@@ -81,4 +83,20 @@ class CdlExtractor extends Extractor with JsonExtractionUtils {
...
+    * Implemented per provider
+    *
+    * DigestUtils.md5Hex(id)
+    *
+    * @return MD5 hash value
+    */
+  override def mintItemId(json: Any): String = {
+    json match {
+      case json: JValue => Utils.generateMd5(Option(List("cdl", extractString("id")(json)).mkString("--")))
+      // TODO: case _ => throw Exception(?) or some other error
```
moltude: Assuming this should throw an error but not sure what kind or type.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
moltude: Calling out this addition to DplaMapData.

src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala
mdellabitta: I think it shouldn't take Any, that seems like too wide open. Maybe it should be oriented toward something like `providerName, stuffToHash` or something?

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
mdellabitta: I think (given my assumption that the IDs are URIs) that this may already be handled in one of the URI fields already there.

mdellabitta: Specifically, the OreAggregation URI

moltude: @mdellabitta I don't think so, `id` is a string but `@id` is a URI (at least according to what I see in ES) 

```
"_source": {
"isShownAt": "http://www.davidrumsey.com/luna/servlet/detail/RUMSEY~8~1~207512~3003388",
"object": "http://www.davidrumsey.com:8480/MediaManager/srvr?mediafile=/Size1/RUMSEY-8-NA/5025/0710_8.jpg",
"id": "0215aeb239de45b0dc789932607a1105",
"sourceResource": {},
"@id": "http://dp.la/api/items/0215aeb239de45b0dc789932607a1105",
"ingestType": "item",
"originalRecord": {}
}```

File: src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala (hunk 3 above)
Comment (mdellabitta) [inline src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala]: I think the best way to do it is you take the extraction of the needed value out of the responsibility of this method, and just make this method take the string and provider (see my comment below).

Comment (mdellabitta) [inline src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala]: To clarify, when I say String, I mean literally the thing to hash, not the whole record.

File: src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala (hunk 1 above)
Comment (moltude) [inline src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala]: FYI to all -- This is tied to the conversation below about what the type of `id` should be (`String` vs `URI`)

Comment (moltude) [inline src/main/scala/dpla/ingestion3/mappers/providers/CdlExtractor.scala]: @mdellabitta Yes to making it its own function

File: src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala (hunk 3 above)
Comment (moltude) [inline src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala]: @mdellabitta The issue I see with that is that we will then need to repeat the ID selection code when minting the `id` value and the `uri` and that leaves us vulnerable to not updating how the id is selected in all places.

File: src/main/scala/dpla/ingestion3/model/DplaMapData.scala (hunk 2 above)
Comment (moltude) [inline src/main/scala/dpla/ingestion3/model/DplaMapData.scala]: @mdellabitta I may be doing this wrong my looking at records in ES...

Comment (mdellabitta) [inline src/main/scala/dpla/ingestion3/model/DplaMapData.scala]: Hmm, I'm basing it off of the DPLA MAP spec. I agree if there needs to be a standalone ID, we need a spot for it.

File: src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala (hunk 3 above)
Comment (mdellabitta) [inline src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala]: That might be, but I think having the method take Any opens us up to a different set of problems as well. In theory, you could stick anything in there (database connection? file?) and not get any help in knowing that you did the wrong thing.

I think we should use the template pattern here: https://en.wikipedia.org/wiki/Template_method_pattern

Have methods in Extractor that can return the ID or the URI, and have an abstract method that knows how to return the ID?

**Comment by moltude** [comment]
@mdellabitta bumping for discussion tomorrow at weekly meeting

**Comment by moltude** [comment]
@mdellabitta fixed up that CDL `@id` mapping issue.

File: src/main/scala/dpla/ingestion3/model/DplaMapData.scala [hunk 5]
```diff
@@ -31,7 +30,10 @@ object DplaMapData {
 case class DplaMapData(
                          sourceResource: DplaSourceResource,
                          edmWebResource: EdmWebResource,
-                         oreAggregation: OreAggregation
+                         oreAggregation: OreAggregation,
+                         // Add id property at the root level (following ES schema)
+                         id: ExactlyOne[String],
+                         `@id`: ExactlyOne[URI]
```
mdellabitta: Is there something that this is called in the DPLA MAP spec that would be more suitable? I don't think @id means anything to anyone aside from Elasticsearch.

moltude:
@mdellabitta removed id and `@id` and fixed up tests. Should be ready for merge after travis and codacy

PR #45, opened 2017-07-26
Generalized file configuration for harvests

moltude:
Refactor the extraction of OAI parameters from command line arguments to configuration files. The OaiHarvster requires one command line argument to a file that contains the configuration values for the a given provider (e.g. `--Dconfig.{file,url,resource}=<path to config>`). This file will include all configurations for a provider and is not limited to harvest properties.

This PR includes a sample OAI configuration file and a complete harvest profile for Digital Commonwealth. The profile includes an additional property,`type`, which will be used to specify the type of harvester to invoke (OAI, API, Primo, File etc.)

Examples:
* By default it will use application.{conf,properties,json} on the classpath but that can be overridden by specifying these system properties, `config.{url, file, resource}`.
* To use an external config file from S3 when running from IntelliJ or other IDE
`-Dconfig.url=https://s3.amazonaws.com/dpla-sw/application.conf` as a VM option
* To use an external configuration file on the filesystem when running from IntelliJ or other IDE
`-Dconfig.file=/home/scott/i3/profiles/provider-cdl.conf`

TODO: 
* Figure out why invocation using SBT and spark-shell is not respecting `config.{url,file,resource}` when passed in at runtime. ee

### DIFF
--- a/src/main/scala/dpla/ingestion3/confs/OaiHarvesterConf.scala
+++ b/src/main/scala/dpla/ingestion3/confs/OaiHarvesterConf.scala
@@ -1,13 +1,11 @@
 package dpla.ingestion3.confs
 
-import org.rogach.scallop.{ScallopConf, ScallopOption}
-import dpla.ingestion3.utils.Utils.validateUrl
+import com.typesafe.config.ConfigFactory
 
 /**
-  * https://github.com/scallop/scallop/wiki
   *
   * Arguments to be passed into the OAI-PMH harvester
-  *
+  * 
   * Output directory: String
   * Endpoint: String the OAI URL
   * Verb: String ListRecords || ListSets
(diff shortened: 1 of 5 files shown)


moltude:
rebased develop to pull over @mdellabitta's date enrichment fix

src/main/scala/dpla/ingestion3/confs/OaiHarvesterConf.scala
```diff
@@ -22,43 +20,53 @@ import dpla.ingestion3.utils.Utils.validateUrl
...
-    default = Some("local[*]"))
+class OaiHarvesterConf(arguments: Seq[String]) {
+  def load(): OaiConfig = {
+    // Loads config file for more information about loading hierarchy please read
+    // https://github.com/typesafehub/config#standard-behavior
+    ConfigFactory.invalidateCaches()
+    val oaiConf = ConfigFactory.load()
+
+    def getProp(prop: String, default: Option[String] = None): Option[String] = {
+      oaiConf.hasPath(prop) match {
```
mdellabitta: I'm wondering if there's a possibility to add environment variable overriding here.

markbreedlove:
Could there be some documentation in README.md about how to invoke the harvester with the necessary commandline argument to read in the configuration?

moltude:
@markbreedlove Will do re: adding documentation to invoke using the provided sample config files

src/main/scala/dpla/ingestion3/confs/OaiHarvesterConf.scala
mdellabitta: That part is built in, right? What I'm suggesting is the possibility of ad hoc overriding a property with an environment variable named the same thing, i.e., `export oaiParams.endpoint=http://example.com && sbt run`

We had talked about that as an easy way to achieve what @markbreedlove  was asking for without having to write a lot of argument parsing code alongside the config code.

PR #46, opened 2017-07-26
Update Avro schemas to match specs, load from files

moltude:
Update Avro schema files to match the specification at
https://digitalpubliclibraryofamerica.atlassian.net/wiki/display/TECH/Ingestion+3+Storage+Specification

Remove hard coded schemas from driver classes.

Add a basic file reader method to FlatFileIO which returns the file
contents as a String.


src/main/scala/dpla/ingestion3/OaiHarvesterMain.scala
```diff
@@ -2,144 +2,89 @@ package dpla.ingestion3
...
-      "harvestAllSets" -> oaiParams.harvestAllSets,
-      "setlist" -> oaiParams.setList,
-      "blacklist" -> oaiParams.blacklist,
-      "endpoint" -> oaiParams.endpoint
+      "verb" -> oaiConf.verb.toOption,
+      "metadataPrefix" -> oaiConf.prefix.toOption,
+      "harvestAllSets" -> oaiConf.harvestAllSets.toOption,
+      "setlist" -> oaiConf.setlist.toOption,
+      "blacklist" -> oaiConf.blacklist.toOption
     ).collect{ case (key, Some(value)) => key -> value } // remove None values
```
mdellabitta: Not part of this PR, so ignore, but this line can be replaced with a call to .flatten() or .flatMap() I think

src/main/scala/dpla/ingestion3/utils/FileIO.scala
```diff
@@ -23,6 +25,16 @@ class FlatFileIO extends FileIO {
     val outFile = new File(outputFile).toPath
     Files.write(outFile, record.getBytes("utf8"), CREATE, TRUNCATE_EXISTING)
   }
+
+  /**
+    * Reads a file and returns it as a single string
+    * @param name
+    * @return
+    */
+  def readFileAsString(name: String): String = {
```
mdellabitta: Pretty sure you need to close the stream.

PR #47, opened 2017-07-26
Add "MVP" spatial enrichment

markbreedlove:
See https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1307

This changeset fills in the spatial enrichment, resolving more fields and refactoring it for various reasons, including for its ability to be tested.

Commit 680d2e2 enables integration tests per what appears to be the standard paradigm. There are just a few integration tests to help keep an eye on Twofishes, and verify behavior that's delegated to Twofishes, like whether it resolves dotted state abbreviations; but most of the tests are unit tests.

`EnricherMain` has some code commented out because it's due for a refactor following the recent addition of the `EnrichmentDriver` class.

`dpla.ingestion3.model.EdmPlace` has been renamed to `DplaPlace` because it's been replaced with the MAP3.1 form of Place.  We're in between MAP3.1 and 4.x here, because MAP4.0 wasn't working out for spatial data.

The `SpatialEnrichmentTest` test suite might be the first place you want to look to see what the new enrichment is doing.


src/main/scala/dpla/ingestion3/enrichments/EnrichmentDriver.scala
```diff
@@ -2,10 +2,25 @@ package dpla.ingestion3.enrichments
...
 
+/**
+  * TODO: assign the hostname from a config file or commandline switch?
+  *
+  * @see dpla.ingestion3.enrichments.Twofisher
+  * @see SpatialEnrichmentIntegrationTest
+  */
+object Geocoder extends Twofisher {
+  override def hostname = {
+    System.getenv("GEOCODER_HOST") match {
```
mdellabitta: At some point, we're going to need to figure out a better way to inject configuration. It also will have to deal with a distributed setup.

markbreedlove: Yes. I hope this is easy enough to override when there's a better way.

markbreedlove:
In classic fashion, I realized there was an oversight after issuing the PR, and have added a commit.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
```diff
@@ -0,0 +1,275 @@
...
+    * it if it's already defined.
+    *
+    * @param nameCandidate
+    * @param j
+    */
+  private def nameUnlessCoords(nameCandidate: Option[String], j: JValue):
+      Option[String] = {
+    nameCandidate match {
+      case s if s.isDefined => {
+        if (s.mkString matches """[\+\-\d\.]+\s*,\s*[\+\-\d\.]+""") {
```
markbreedlove: This only catches decimal coordinates. It won't work with degrees-minutes-seconds notation. It may be best to search for any digits, because they don't usually show up in place names. The damage would be very low if the few false positives got overwritten with the names coming out of Twofishes.

src/main/scala/dpla/ingestion3/enrichments/EnrichmentDriver.scala
mdellabitta: Yeah, that's a tidy way to get around it.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  /**
+    * Return the place's display name, for sourceResource.spatial.name
+    *
+    * @param j  The JValue object for the Twofishes response body
+    * @return   The place's display name
+    */
+  private def displayName(j: JValue): Option[String] = {
+    Option(((j \ "interpretations")(0) \ "feature")
```
mdellabitta: It seems like, if the thing you're passing in feels like "something", it becomes a Some, but if it's a JNull or JNothing or something, it becomes a None.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
```diff
@@ -0,0 +1,275 @@
...
+  /**
+    * Return a single String of the Place's coordinates, joined by ","
+    *
+    * @param j  Twofishes response JValue
+    * @return   The coordinates as at "lat, lng" string
+    */
+  private def coordinates(j: JValue): Option[String] = {
+    val c = ((((j \ "interpretations")(0) \ "feature") \ "geometry") \
+      "center").extract[Coords]
+    Option(c.lat.toString + "," + c.lng.toString)
```
mdellabitta: Here I would say this is definitely a Some.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
```diff
@@ -0,0 +1,275 @@
...
+    * to override this when we fill it in.  Otherwise, we don't want to clobber
+    * it if it's already defined.
+    *
+    * @param nameCandidate
+    * @param j
+    */
+  private def nameUnlessCoords(nameCandidate: Option[String], j: JValue):
+      Option[String] = {
+    nameCandidate match {
+      case s if s.isDefined => {
```
mdellabitta: Here I would say `case Some(value)` and let the pattern match destructure the Option for you.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
mdellabitta: Here, you wouldn't use `s.mkString` if you used my previous comment, you'd just use `value matches...`

I believe there's even a way to get pattern matching to work with regular expressions, but I haven't messed with it myself.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
```diff
@@ -0,0 +1,275 @@
...
+  /**
+    * Enrich the given DplaPlace
+    *
+    * @param p  The Place
+    * @return   An enriched copy of the Place
+    */
+  def enrich(p: DplaPlace): DplaPlace = {
+    try {
+      queryTerm(p) match {
+        case Some(term) => {
```
mdellabitta: Trivia: you don't need curly braces to wrap a group of statements under a case statement.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
```diff
@@ -0,0 +1,275 @@
...
+            )
+          } else {
+            p
+          }
+        }
+        case _ => p
+      }
+    } catch {
+      case e: Exception => {
+        logger.error(e.toString)  // TODO: Catch this higher up and store it?
```
mdellabitta: Do we know what types of errors might happen? I think we want to distinguish the cases of "twofishes blew up" from the cases of "couldn't find a result for x". And we also don't want to have to check the logs to find out that our output is incorrect because of a network problem. I'd argue for actually letting the exception bubble up here.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
```diff
@@ -0,0 +1,275 @@
...
+    * @param p  The Place
+    * @return   An enriched copy of the Place
+    */
+  def enrich(p: DplaPlace): DplaPlace = {
+    try {
+      queryTerm(p) match {
+        case Some(term) => {
+          val j = geocoder.geocoderResponse(term)
+          // Return the looked-up place, unless it was a country.
+          if (woeType(j) != Some(12)) {
```
moltude: can the variables be renamed to something easier to follow (originalPlace, enrichedPlace)?  I can see what is happening here but the lines below (253-259) are hard to follow with just `p` and `j`

markbreedlove:
Sorry to do this again, but in addition to attending to the things you all pointed out, I have to make this thing do reverse lookups (where you look up a place based on coordinates, not its name). I left that out! I'll have to make some more commits before review on this PR proceeds.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
markbreedlove: It wasn't intuitive at first, and I think I got to this by looking at an example somewhere else. It looks like the return value has to be explicitly wrapped in this Option wrapper to match the declared return type.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
markbreedlove: `Some` works (though the code works the same either way, according to the unit tests). I've changed it & will commit soon.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
markbreedlove: Thanks, just encountered that when looking over IntelliJ IDEA's suggestions. Will commit this soon.

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
mdellabitta: Yeah, I looked it up. The Option(value) version is for if the value might be null, It returns None if null, Some(value) otherwise. So it promotes a type from a nullable type to a Option type that's not nullable.

Some(value) for something that's nullable will return Some(null) in those cases, which is usually not intended.

mdellabitta: Oh cool, I just figured out how to fix a Codacy error with this!

src/main/scala/dpla/ingestion3/enrichments/SpatialEnrichment.scala
markbreedlove: I didn't know enough about the different types of exceptions to do more than catch the `Exception` base class. I did try to use a `Try` instead of this older-style try/catch block, but couldn't get it to work in the time I wanted to spend.

Yes, I was attempting to indicate from the comment that I think we might want to have this bubble up and store it with the result set, but I'd prefer to send it simultaneously to the logger. I wouldn't want to have to run a Spark job or other analytics / QA job to get the results of error messages that could be related to system events.  I'd want the errors in both places. If logged to syslog they could show up almost instantly in whatever we're using to monitor other parts of the system. If someome runs a sudo command on server A that knocks out Twofishes on server B, such that the enrichment job throws an exception, we could see all the related messages together in one place with one u.i. like Kibana. The logger doesn't have to be seen as creating log files -- it can log in various ways.

mdellabitta: If you let this exception propagate, it'll end up at the driver for the operator to see, actually. If you only log it, you have to go to a log file/system to find out that anything is wrong.

At the very least, you should rethrow the exception with `throw e` I think.

markbreedlove: I was thinking it would either be logged and rethrown, or be caught higher up and then logged and persisted. I wouldn't want to just log it and not persist it.

mdellabitta: I agree, but the safest way to leave this if you haven't done that is to have it throw for now, so we're forced to do something about it, rather than our forgetting about it and having it bite us later, was my thinking.

markbreedlove: Oh, I see. I was catching it because there's currently no mechanism to persist the errors, thus the comment "Catch this higher up and store it?" If you like I'll just take out the try/catch.

mdellabitta: Well, that gets rid of your log statement. Either way...

markbreedlove: Hopefully we can deal with this soon, both logging and persisting the errors upstream.

build.sbt
```diff
@@ -1,38 +1,38 @@
...
-  "org.jsoup" % "jsoup" % "1.10.2",
-  "org.scalaj" %% "scalaj-http" % "2.3.0",
-  "org.eclipse.rdf4j" % "rdf4j-model" % "2.2",
-  "org.eclipse.rdf4j" % "rdf4j-rio-api" % "2.2",
-  "org.eclipse.rdf4j" % "rdf4j-rio-turtle" % "2.2",
-  "org.scalaj" % "scalaj-http_2.11" % "2.3.0",
-  "org.rogach" % "scallop_2.11" % "3.0.3",
-  "org.scalamock" %% "scalamock-scalatest-support" % "3.6.0" % "test"
-)
+lazy val root = (project in file("."))
```
AudreyAltman: Could you add a comment here about why we need the lazy evaluation?  I imagine that in six months time I will have forgotten!

mdellabitta: I bet @markbreedlove was following code recommended by sbt. I wonder if doing it lazily means it's compatible with codegen or something.

markbreedlove: Yes, I was emulating some code that `sbt init` (I think) produced. I don't know its significance. Need to spend more time w Scala ...

markbreedlove:
Reviewers, please note the addition of two more commits that fix a bug in which an unsuccessful lookup resulted in an unhandled exception that we wouldn't want to have bubble up.

markbreedlove:
Note also that the integration tests run successfully now, but were not before 694f5d5 .

PR #48, opened 2017-07-27
Fixed nondeterministic execution

mdellabitta:
Changed the way fallthrough execution works in parse() to be cleaner
and to avoid race conditions.


src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
```diff
@@ -107,17 +99,44 @@ class ParseDateEnrichment {
...
-          Try(head.parse(str)) match {
-            case Success(date) => Some(responseFormat.format(date))
-            case Failure(exception) => innerParseDate(str, rest)
+          val df = new SimpleDateFormat(head)
+          df.setLenient(false)
+          val responseFormat = new SimpleDateFormat("yyyy-MM-dd")
+          df.setLenient(false)
+          Try(df.parse(str)) match {
+            case Success(date) =>
+              println(head + " matched against " + str + " and yielded " + responseFormat.format(date))
```
moltude: Remove debugging print statements?

moltude:
@mdellabitta those codacy errors are a big `¯\_(ツ)_/¯` from me. Doesn't seem like there is a way around using `return` in this case.

PR #50, opened 2017-07-28
NARA Mapper first pass

mdellabitta:
There's a few TODOs and things I'll need to talk to Gretchen about. 
Ran this locally on my laptop and it was done in well under an hour.
Codacy will probably yell at me, and I'll work on what it says.

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
@@ -1,0 +1,255 @@
+package dpla.ingestion3.mappers.providers
+
+import java.net.URI
+
+import dpla.ingestion3.enrichments.{DcmiTypeMapper, VocabEnforcer}
+import dpla.ingestion3.mappers.rdf.DCMIType
+import dpla.ingestion3.mappers.xml.XmlExtractionUtils
+import dpla.ingestion3.model._
+import org.eclipse.rdf4j.model.IRI
+
+import scala.xml.{Node, NodeSeq, XML}
+
+class NaraExtractor(rawData:String) extends Extractor with XmlExtractionUtils with Serializable {
+
+  implicit val xml: NodeSeq = XML.loadString(rawData)
+
+  override def getProviderBaseId(): Option[String] = Some("nara--" + itemUri.toString) //TODO can't figure this out.
+
+  override def build(): DplaMapData = {
+    DplaMapData(
+      DplaSourceResource(
+        collection = collection(xml).map(nameOnlyCollection),
+        contributor = contributor(xml).map(nameOnlyAgent),
+        creator = creator(xml).map(nameOnlyAgent),
+        date = date(xml),
+        description = extractStrings("scopeAndContentNote"),
+        extent = extractStrings("extent"),
+        format = extractStrings(xml \\ "specificRecordsTypeArray" \ "specificRecordsType" \ "termName"),
+        identifier = extractStrings("naId"),
+        language = extractStrings(xml \\ "languageArray" \ "language" \ "termName").map(nameOnlyConcept),
+        place =
+          extractStrings(xml \\ "geographicReferenceArray" \ "geographicPlaceName" \ "termName").map(nameOnlyPlace),
+        publisher = publisher(xml).map(nameOnlyAgent),
+        relation = relation(xml).map(Left(_)),
+        rights = rights(xml),
+        subject = extractStrings(xml \\ "topicalSubjectArray" \ "topicalSubject" \ "termName").map(nameOnlyConcept),
+        title = extractStrings("title"),
+        `type` = types(xml)
+      ),
+      EdmWebResource(
+        //todo proper uri handling from other branch
+        uri = itemUri(xml),
+        //TODO naId doesn't seem like a meaningful file format
+        fileFormat = extractStrings(xml \\ "objectType" \ "naId") ++ extractStrings(xml \\ "objectType" \ "termName")
+      ),
+      OreAggregation(
+        uri = itemUri(xml), //this should probably be renamed to isShownAt
+        dataProvider = dataProvider(xml),
+        originalRecord = rawData,
+        provider = agent,
+        preview = extractString("thumbnailFilename").map(new URI(_)).map(uriOnlyWebResource)
+      )
+    )
+  }
+
+  def agent = EdmAgent(
+    name = Some("National Archives and Records Administration"),
+    uri = Some(new URI("http://dp.la/api/contributor/nara"))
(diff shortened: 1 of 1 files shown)


mdellabitta:
Clearly needs a rebase on the date fix.

mdellabitta:
Did a rebase to get the `develop` in there, but didn't squash yet.

src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
```diff
@@ -0,0 +1,256 @@
...
+    f"""$value3"; $value2"; "$value1""""
+  }
+
+  def rights(xml: NodeSeq): Seq[String] = for {
+    useRestriction <- xml \ "useRestriction"
+    value1 = (useRestriction \ "note").text
+    value2 = (useRestriction \ "specificUseRestrictionArray" \ "specificUseRestriction" \ "termName").text
+    value3 = (useRestriction \ "status" \ "termName").text
+  //VALUE2": "VALUE1" "VALUE3
+  } yield f"""$value2": "$value1" "$value3"""
```
markbreedlove: Are these double-quotes correct? They appear not to match the pattern in the comment above.

src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
```diff
@@ -0,0 +1,256 @@
...
+      coverageDate <- xml \\ "coverageDates"
+      coverageStartDate = (coverageDate \ "coverageStartDate").headOption
+      coverageEndDate = (coverageDate \ "coverageEndDate").headOption
+      //dateQualifier = coverageDate \ "dateQualifier" //todo not sure what to do with qualifier in EDTF
+      if coverageStartDate.nonEmpty || coverageEndDate.nonEmpty
+
+    } yield {
+      val displayDate = getDisplayDate(coverageStartDate, coverageEndDate)
+      EdmTimeSpan(
+        originalSourceDate = displayDate, //this is cheating?
```
markbreedlove: `sourceResource.date` is supposed to be for date of publication or date of creation, as I understand it, whereas `sourceResourc.temporal` is for "aboutness." Therefore, with regard to whether this is cheating, I'd just say it's not if you can confirm that these "coverage dates" are for the date the source resource was generated or published.

src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
```diff
@@ -0,0 +1,256 @@
...
+import dpla.ingestion3.mappers.rdf.DCMIType
+import dpla.ingestion3.mappers.xml.XmlExtractionUtils
+import dpla.ingestion3.model._
+import org.eclipse.rdf4j.model.IRI
+
+import scala.xml.{Node, NodeSeq, XML}
+
+class NaraExtractor(rawData:String) extends Extractor with XmlExtractionUtils with Serializable {
+
+  implicit val xml: NodeSeq = XML.loadString(rawData)
```
markbreedlove: Could you comment on why this `implicit val xml` is an implicit val?

I see that `dpla.ingestion3.mappers.xml.XmlExtractionUtils` has the methods `extractString` and `extractStrings`, which each have an `implicit val xml`, but I have the sense I'm missing something when I read this.

mdellabitta: Is there something I should comment or explain?

markbreedlove: I guess just why not make it `val xml`?

src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
mdellabitta: They seem to match to me, but maybe I'm taking the notation too literally?

Note that the triple quotes at the beginning and end mean that single quotes no longer require escaping. Also note that I have no idea what's up with the specification in the comment, and I could be implementing it entirely incorrectly.

src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
mdellabitta: It means I don't need to pass it explicitly to every method, which reduces noise. We've followed this pattern in the other mappers.

src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
markbreedlove: Oh, right, I guess I didn't look at that closely.
Still, as you mentioned, I don't get the unclosed double-quotes in the specification.

src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
```diff
@@ -0,0 +1,256 @@
...
+      org <- xml \\ "organizationalContributorArray" \ "organizationalContributor"
+      name = (org \ "contributor" \ "termName").text
+      _type = (org \ "contributorType" \ "termName").text
+      if _type != "Publisher"
+    } yield name
+
+    val personalContributors = for {
+      person <- xml \\ "personalContributorArray" \ "personalContributor"
+      name = (person \ "contributor" \ "termName").text
+      //_type = (person \ "contributorType" \ "TermName").text
```
AudreyAltman: Do you want to keep the commented-out line above?

src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
```diff
@@ -0,0 +1,256 @@
...
+    val releaseDates = simpleDate(xml \\ "releaseDateArray" \ "proposableQualifiableDate")
+
+    Seq(
+      coverageDates,
+      copyrightDates,
+      productionDates,
+      broadcastDates,
+      releaseDates
+    )
+      .collectFirst({ case x => x })
```
AudreyAltman: I'm not clear what this `collectFirst` method is doing.

mdellabitta: It gets the first result that's not None. So, the first execution that returns a value. There's a better way to do this, though, so I'll check that in instead.

src/main/scala/dpla/ingestion3/mappers/providers/NaraExtractor.scala
mdellabitta: It's mentioned in the spec, but doesn't do anything. Not quite sure what to do with it, to be honest!

PR #51, opened 2017-07-28
Dt 1474 abstract oai class

AudreyAltman:
The harvest does not fail completely due to errors with single requests or responses.  Instead, the harvest will complete as much as possible, and return both successful responses, and actionable information about failures.

New case classes allow us to track information about the harvesting process.  Each individual request to the OAI feed begins with the instantiation of an `OaiSource`, which at minimum stores the parameters for that query.  

1. Query is initiated.  `OaiSource` is instantiated with the query params.
2. URL is formed from the query param and stored in the `OaiSource`.
3. HTTP request is made.  Response is stored in `OaiSource` as a String.
4. Response string is parsed into XML.
5. Response XML is checked to see if it contains an error message.  
6. Response XML is parsed into records or sets.

An error can occur during steps 2-5, in which case an `OaiError` is returned with the `OaiSource`.  Otherwise, records and sets are returned, also with their respective `OaiSource`s.

The new schema for the output DataFrame is:

```
root
 |-- set: struct (nullable = true)
 |    |-- id: string (nullable = true)
 |    |-- document: string (nullable = true)
 |    |-- setSource: struct (nullable = true)
 |    |    |-- queryParams: map (nullable = true)
 |    |    |    |-- key: string
 |    |    |    |-- value: string (valueContainsNull = true)
 |    |    |-- url: string (nullable = true)
 |    |    |-- text: string (nullable = true)
 |-- record: struct (nullable = true)
 |    |-- id: string (nullable = true)
 |    |-- document: string (nullable = true)
 |    |-- setIds: array (nullable = true)
 |    |    |-- element: string (containsNull = true)
 |    |-- recordSource: struct (nullable = true)
 |    |    |-- queryParams: map (nullable = true)
 |    |    |    |-- key: string
 |    |    |    |-- value: string (valueContainsNull = true)
 |    |    |-- url: string (nullable = true)
 |    |    |-- text: string (nullable = true)
 |-- error: struct (nullable = true)
 |    |-- message: string (nullable = true)
 |    |-- errorSource: struct (nullable = true)
 |    |    |-- queryParams: map (nullable = true)
 |    |    |    |-- key: string
 |    |    |    |-- value: string (valueContainsNull = true)
 |    |    |-- url: string (nullable = true)
 |    |    |-- text: string (nullable = true)
```

Note: Sets can no longer be persisted as CSV files b/c they have a complex, nested data type.  They are saved as AVRO files.

### DIFF
--- a/src/main/scala/dpla/ingestion3/OaiHarvesterMain.scala
+++ b/src/main/scala/dpla/ingestion3/OaiHarvesterMain.scala
@@ -15,48 +15,8 @@ import scala.util.{Failure, Success, Try}
 
 /**
   * Entry point for running an OAI harvest.
-  *
-  * For argument options, @see OaiHarvesterConf.
   */
 object OaiHarvesterMain {
-
-  val recordSchemaStr =
-    """{
-        "namespace": "dpla.avro",
-        "type": "record",
-        "name": "OriginalRecord.v1",
-        "doc": "",
-        "fields": [
-          {"name": "id", "type": "string"},
-          {"name": "document", "type": "string"},
-          {"name": "set_id", "type": "string"},
-          ("name": "set_document", "type": "string"},
-          {"name": "provider", "type": "string"},
-          {"name": "mimetype", "type": { "name": "MimeType",
-           "type": "enum", "symbols": ["application_json", "application_xml", "text_turtle"]}
-           }
-        ]
-      }
-    """//.stripMargin // TODO we need to template the document field so we can record info there
-
-  // This schema String is printed to help with debugging.
-  // It is NOT implemented during the write operation b/c sets are written to CSV.
-  val setSchemaStr =
-    """{
-        "namespace": "dpla.avro",
-        "type": "set",
-        "name": "OriginalRecord.v1",
-        "doc": "",
-        "fields": [
-          {"name": "id", "type": "string"},
-          {"name": "document", "type": "string"},
-          {"name": "provider", "type": "string"},
-          {"name": "mimetype", "type": { "name": "MimeType",
-           "type": "enum", "symbols": ["application_json", "application_xml", "text_turtle"]}
-           }
-        ]
-      }
-    """
 
   val logger = LogManager.getLogger(OaiHarvesterMain.getClass)
 
@@ -106,7 +66,7 @@ object OaiHarvesterMain {
     }
 
     runHarvest() match {
-      case Success(results) => {
+      case Success(results) =>
         results.persist(StorageLevel.DISK_ONLY)
 
         val dataframe = results.withColumn("provider", lit(provider))
@@ -115,28 +75,13 @@ object OaiHarvesterMain {
         // Log the results of the harvest
         logger.info(s"Harvested ${dataframe.count()} records")
 
-        readerOptions("verb") match {
-          // Write records to avro.
-          // This task may require a large amount of driver memory.
-          case "ListRecords" => {
-            println(recordSchemaStr)
+        val schema = dataframe.schema
 
-            dataframe.write
-              .format("com.databricks.spark.avro")
-              .option("avroSchema", recordSchemaStr)
-              .avro(outputDir)
-          }
-          // Write sets to csv.
-          case "ListSets" => {
-            println(setSchemaStr)
+        dataframe.write
+          .format("com.databricks.spark.avro")
+          .option("avroSchema", schema.toString)
+          .avro(outputDir)
 
-            dataframe.coalesce(1).write
-              .format("com.databricks.spark.csv")
-              .option("header", true)
-              .csv(outputDir)
-          }
-        }
-      }
       case Failure(f) => logger.fatal(s"Unable to harvest records. ${f.getMessage}")
     }
     // Stop spark session.
--- a/src/main/scala/dpla/ingestion3/harvesters/oai/OaiRelation.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/oai/OaiRelation.scala
@@ -4,6 +4,7 @@ import org.apache.spark.sql._
 import org.apache.spark.sql.sources._
 import org.apache.spark.sql.types._
 import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.catalyst.ScalaReflection
 
 /**
   * This class interprets the user-submitted params to determine which type of
@@ -32,13 +33,13 @@ class OaiRelation (endpoint: String,
 
   val oaiResponseBuilder = new OaiResponseBuilder(endpoint)(sqlContext)
 
-  def allSets: RDD[OaiSet] = oaiResponseBuilder.getSets
+  def allSets: RDD[OaiResponse] = oaiResponseBuilder.getAllSets
 
   /**
     * Make appropriate call to OaiResponseBuilder based on presence or absence of
     * set, blacklist, or harvetAllSets.
     */
-  def records: RDD[OaiRecord] = {
+  def records: RDD[OaiResponse] = {
 
     /**
       * Get any optional OAI arguments that may be used in a records request.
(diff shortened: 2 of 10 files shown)


src/main/scala/dpla/ingestion3/OaiHarvesterMain.scala
```diff
@@ -95,28 +56,13 @@ object OaiHarvesterMain {
...
-
-        dataframe.coalesce(1).write
-          .format("com.databricks.spark.csv")
-          .option("header", true)
-          .csv(oaiConf.outputDir())
-      }
-      case _ => throw new IllegalArgumentException("Verb not recognized.")
-    }
+    val schema = dataframe.schema
+    println(dataframe.printSchema())
```
mdellabitta: Assuming this is debug code that should be removed?

src/main/scala/dpla/ingestion3/harvesters/oai/OaiRelation.scala
```diff
@@ -55,73 +56,97 @@ class OaiRelation (endpoint: String,
...
-          StructField("document", StringType, true)))
+  /**
+    * Get any relevant sets for a records request, based on presence of setlit,
+    * blacklist, or harvestAllSets.
+    *
+    * @return all sets from which records should be harvested, and any errors
+    *         incurred in the process of harvesting said sets
+    */
+  def setResponses: Option[RDD[OaiResponse]] = {
+    (setlist, blacklist, harvestAllSets) match {
```
mdellabitta: Nice use of this!

src/main/scala/dpla/ingestion3/harvesters/oai/OaiRelation.scala
```diff
@@ -55,73 +56,97 @@ class OaiRelation (endpoint: String,
...
+    * For example, if a record belongs to more than one set, it may appear in
+    * responses for each set.
+    */
+  override def buildScan(): RDD[Row] = {
+
+    getResponses.flatMap {
+      case x: SetsPage => x.sets.map(set => Row(set, None, None))
+      case x: RecordsPage => x.records.map(record => Row(None, record, None))
+      case x: OaiError => Seq(Row(None, None, x))
+      case x: OaiSource => {
```
mdellabitta: You don't need this level of curly braces just FYI. `case foo =>` statements specify their own block scopes without the braces.

src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseBuilder.scala
```diff
@@ -21,95 +24,148 @@ class OaiResponseBuilder (endpoint: String)
...
+      // Apply partial function to process the OaiSource in some specific way.
+      case page: OaiSource => pageProcessor(page)
+      // Retain other responses (ie. errors) as given.
+      case response: OaiResponse => response
+    }
+  }
+
+  def getAllSets(): RDD[OaiResponse] = {
+    val allSetsPf: PartialFunction[(OaiSource), OaiResponse] = {
+      case(page) => OaiResponseProcessor.getAllSets(page)
```
mdellabitta: Not familiar with this `case(blah)` syntax, is that different than `case blah:`?

src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
```diff
@@ -18,40 +19,59 @@ object OaiResponseProcessor {
...
-      records.map(r => r.headOption match {
-        case Some(node) => {
-          new OaiRecord(getOaiIdentifier(node), node.toString, set)
+  def getRecords(source: OaiSource): OaiResponse = {
+
+    // Parse all records from XML
+    def getRecordsFromXml(xml: Node): Seq[OaiRecord] = {
+      val xmlRecords: NodeSeq = xml \\ "OAI-PMH" \\ "record"
+
+      xmlRecords.flatMap(record => {
```
mdellabitta: You don't need curly braces here, either, since `record.headOption match {}` is a single statement

src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
```diff
@@ -60,20 +80,102 @@ object OaiResponseProcessor {
     * @return Seq[String]
     *         The set ids.
     */
-  def getSetIdsFromRecord(xml: NodeSeq): Seq[String] = {
+  def getSetIdsFromRecord(xml: Node): Seq[String] = {
     val sets = xml \\ "setSpec"
```
mdellabitta: Possibly more idiomatic: `for (set <- xml \\ "setSpec") yield set.text`

src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
```diff
@@ -60,20 +80,102 @@ object OaiResponseProcessor {
...
+    *                   A partial function indicating how sets should be filtered,
+    *                   ie. according to a whitelist or blacklist.
+    */
+  def getSets(source: OaiSource, setFilter: PartialFunction[(Seq[OaiSet]), Seq[OaiSet]]): OaiResponse = {
+
+    // Parse all sets from XML.
+    def getSetsFromXml(xml: Node): Seq[OaiSet] = {
+      val xmlSets = xml \\ "OAI-PMH" \\ "set"
+
+      xmlSets.flatMap(set => set.headOption match {
```
mdellabitta: This might be cleaner with a `for`, too, and you wouldn't need the pattern match. Winging it in github:

`for (set <- xml \\ "OAI-PMH" \\ "set") yield Some(OaiSet(getSetIdentifier(set), node.text, source))`

Also, if these nodes are definitely children of one another, using `\` is preferred because `\\` means you want it to dig through all the descendants, not just the first level.

src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
```diff
@@ -60,20 +80,102 @@ object OaiResponseProcessor {
...
+  }
+
+  /**
+    * Accepts a set from an OAI feed an returns the OAI identifier
+    *
+    * @param set Node
+    *            The original set from the OAI feed
+    * @return The local identifier
+    */
+  def getSetIdentifier(set: Node): String = {
```
mdellabitta: I'd probably eliminate the curly braces for one liners like this.

src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
```diff
@@ -60,20 +80,102 @@ object OaiResponseProcessor {
...
+    *             The XML that may include an error message.
+    *
+    * @return Try[Unit]
+    *
+    * @throws OaiHarvesterException
     *
     */
-  @throws(classOf[OaiHarvesterException])
-  def getOaiErrorCode(xml: NodeSeq): Unit = {
+  def getOaiErrorCode(xml: NodeSeq): Try[Unit] = Try{
```
mdellabitta: This would probably be cleaner as an if, although I wonder if Codacy would complain! :)

```
if ((xml \\ "OAI-PMH" \\ "error").nonEmpty)  
  throw new OaiHarvesterException((xml \\ "OAI-PMH" \\ "error").text.trim)
```

src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseBuilder.scala
AudreyAltman: Nope, not different - I'll clean it up, thanks!

src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
AudreyAltman: @moltude maybe you could lend your OAI expertise here - is it safe to assume that `"OAI-PMH"` is a direct descendent of the root xml, and/or that `"set"` is a direct descendent of `"OAI-PMH"`?

src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
AudreyAltman: I'm not going to have any curly braces left! ;)  If I were a dragon, I would be jealously guarding a gleaming pile of golden curly braces.

PR #53, opened 2017-08-01
Update Avro schemas to match specs, load from files

moltude:
Update Avro schema files to match the specification at
https://digitalpubliclibraryofamerica.atlassian.net/wiki/display/TECH/Ingestion+3+Storage+Specification

Remove hard coded schemas from driver classes.

Add a basic file reader method to FlatFileIO which returns the file
contents as a String.


mdellabitta:
Not particularly worried about the Codacy warning, but feel free if you wanna.

PR #55, opened 2017-08-04
Add mapping and enrichment entry points

moltude:
Adds entry points for invoking mapping (a.k.a. `Extractors`) and enrichment jobs. Both entries accept only two parameters, the source data and output destination. 

The mapping entry point determines the correct mapper extractor to invoke by inspecting the `provider` property stored with the harvested data. 

The enrichment entry point only invokes the enrichments which are applied to all providers (see EnrichmentDriver for a complete list). Make all enrichment classes `Serailizable` to allow distribution over nodes.

**Running via sbt** 
`sbt "run-main dpla.ingestion3.MappingEntry /input/path/to/harvestedData.avro /output/path/to/mappedData.avro"`

`sbt "run-main dpla.ingestion3.EnrichEntry /input/path/to/mappedData.avro /output/path/to/enrichedData.avro"`

This also adds a retry() method to support geocodeResponse(). Tests showed that a single failed request to Twofishes, such as a timeout,  would fail the entire enrichment job. The geocodeResponse() method will now attempt to retry a failed request. If it does not succeed it will return a JNothing object which will cause the original value to be returned by #enrich().

### DIFF
--- a/src/main/scala/dpla/ingestion3/EnrichEntry.scala
+++ b/src/main/scala/dpla/ingestion3/EnrichEntry.scala
@@ -1,0 +1,89 @@
+package dpla.ingestion3
+
+import java.io.File
+
+import dpla.ingestion3.enrichments.EnrichmentDriver
+import dpla.ingestion3.model.DplaMapData
+import dpla.ingestion3.utils.Utils
+import org.apache.log4j.LogManager
+import org.apache.spark.SparkConf
+import org.apache.spark.sql.SparkSession
+
+/**
+  * Expects two parameters:
+  *   1) a path to the harvested data
+  *   2) a path to output the mapped data
+  *
+  *   Usage
+  *   -----
+  *   To invoke via sbt:
+  *     sbt "run-main dpla.ingestion3.EnrichEntry /input/path/to/mapped.avro /output/path/to/enriched.avro"
+  *
+  */
+
+object EnrichEntry {
+
+  def main(args: Array[String]): Unit = {
+
+    val logger = LogManager.getLogger(EnrichEntry.getClass)
+
+    if (args.length != 2)
+      logger.error("Incorrect number of parameters provided. Expected <input> <output>")
+
+    // Get files
+    val dataIn = args(0)
+    val dataOut = args(1)
+
+    val sparkConf = new SparkConf()
+      .setAppName("Enrichment")
+      // TODO there should be a central place to store the sparkMaster
+      .setMaster("local[*]")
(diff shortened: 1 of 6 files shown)


src/main/scala/dpla/ingestion3/EnrichEntry.scala
```diff
@@ -0,0 +1,88 @@
...
+    implicit val dplaMapDataEncoder = org.apache.spark.sql.Encoders.kryo[DplaMapData]
+
+    val spark = SparkSession.builder()
+      .config(sparkConf)
+      .getOrCreate()
+
+    val sc = spark.sparkContext
+
+    // Need to keep this here despite what IntelliJ and Codacy say
+    import spark.implicits._
```
moltude: `spark` in this case refers to the spark session specifically and not a static library that can be imported. I don't fully understand it but that is the gist of what is going on.

src/main/scala/dpla/ingestion3/EnrichEntry.scala
```diff
@@ -0,0 +1,88 @@
...
+    // Save mapped records out to Avro file
+    enrichedRecords.toDF("document").write
+      .format("com.databricks.spark.avro")
+      .save(dataOut)
+
+    sc.stop()
+
+    // Gather some stats
+    val mappedRecordCount = mappedRecords.count()
+    val enrichedRecordCount = enrichedRecords.count()
```
moltude: Crap you are right about this. I made that change during code cleanup. I'll move it up above `sc.stop`

src/main/scala/dpla/ingestion3/EnrichEntry.scala
AudreyAltman: ahhh that makes sense

src/main/scala/dpla/ingestion3/EnrichEntry.scala
```diff
@@ -0,0 +1,88 @@
...
+
+    // Create a broadcast enrichment driver
+    val enrichment = sc.broadcast(new EnrichmentDriver)
+
+    // Run the enrichments over the Dataframe
+    val enrichedRecords = mappedRecords.map(
+      record => {
+        enrichment.value.enrich(record)
+      }
+    )
```
moltude: @AudreyAltman I tried implementing this but ran into some serialization errors. Can we put a pin in this change and revisit it in a future ticket?

AudreyAltman: I was afraid that might happen!  Yeah, we can revisit later.

PR #56, opened 2017-08-04
Add QA report functionality, Property Value Report

markbreedlove:
* Add ReporterMain, Reporter, Report, and PropertyValueReport

See https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1499

Please see comments in the code related to testing and design patterns, and note FIXMEs and TODOs.

### DIFF
--- a/src/main/scala/dpla/ingestion3/ReporterMain.scala
+++ b/src/main/scala/dpla/ingestion3/ReporterMain.scala
@@ -1,0 +1,98 @@
+package dpla.ingestion3
+
+import org.apache.log4j.{Logger, LogManager}
+import dpla.ingestion3.reports._
+import util.{Try, Failure}
+
+
+/*
+ * ReporterMain and Reporter, for generating QA reports
+ *
+ * Example invocation:
+ *
+  *   $ sbt "run-main dpla.ingestion3.ReporterMain \
+  *     /path/to/enriched-data.avro /path/to/propvalreport local[2] \
+  *     propertyValue sourceResource.type"
+ */
+
+
+/**
+  * Reporter, the report driver class.
+  *
+  * This is a class that is instantiated by ReporterMain because the original
+  * idea was it would be possible to unit test this way.  Note that getReport()
+  * is probably the only method that is worth testing.  See the unit tests in
+  * ReporterTest.
+  *
+  * @see ReporterTest
+  *
+  * @param token            The report name token
+  * @param inputURI         Input URI or file path
+  * @param outputURI        Output URI or file path
+  * @param sparkMasterName  Spark master name, e.g. "local[1]"
+  * @param reportParams     Additional parameters particular to the report
+  */
+class Reporter (
+                 token: String,
+                 inputURI: String,
+                 outputURI: String,
+                 sparkMasterName: String,
+                 reportParams: Array[String]
+               ) {
+
+  val logger: Logger = LogManager.getLogger(this.getClass)
+
+  private def getReport(token: String): Option[Report] = {
+    token match {
+      case "propertyValue" =>
+        Some(new PropertyValueReport(
+          inputURI, outputURI, sparkMasterName, reportParams
+        ))
+      case _ => None
+    }
+  }
+
+  def main(): Unit = {
+    val reportResult: Try[Unit] = getReport(token) match {
+      case Some(report) => report.run()
+      case None => Failure(
+        new RuntimeException(s"Report type $token is unknown")
+      )
+    }
+    reportResult match {
+      case Failure(ex) =>
+        logger.error(ex.toString)
+        logger.error("\n" + ex.getStackTrace.mkString("\n"))
+      case _ => Unit
+    }
+  }
+}
+
+
+object ReporterMain {
+
+  def usage(): Unit = {
+    println(
+      """
+        |Usage:
+        |
+        |ReporterMain <input> <output> <spark master> <report token> \
+        |            [<param> ...]
+      """.stripMargin)
+  }
+  def main(args: Array[String]): Unit = {
+    if (args.length < 4) {
+      usage()
+      System.err.println("Incorrect invocation arguments")
+      sys.exit(1)
+    }
+    val inputURI = args(0)
+    val outputURI = args(1)
+    val sparkMasterName = args(2)
+    val token = args(3)
+    val reportParams = args.slice(4, args.length)
+    new Reporter(
+      token, inputURI, outputURI, sparkMasterName, reportParams
+    ).main()
+  }
+}
--- a/src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
+++ b/src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
@@ -1,0 +1,108 @@
+package dpla.ingestion3.reports
+
+import org.apache.spark.sql.{DataFrame, Row, SparkSession, Dataset}
+import dpla.ingestion3.model._
+
+/**
+  * Property Value QA report.  Takes one field and gives a count of unique
+  * values.
+  *
+  * Fields currently supported:
+  *   - sourceResource.format
+  *   - sourceResource.rights
+  *   - sourceResource.type
+  *
+  * @example Example invocation:
+  *
+  *   $ sbt "run-main dpla.ingestion3.ReporterMain \
+  *     /path/to/enriched-data.avro /path/to/propvalreport local[2] \
+  *     propertyValue sourceResource.type"
+  *
+  *     ... This produces a directory named 'propvalreport' with a file in it
+  *     named 'part-< ... >.csv', with contents like:
+  *
+  *     value,count
+  *     physicalobject,1
+  *     dataset,13
+  *     [...]
+  *
+  * @param inputURI         Input URI or file path (Avro file / directory)
+  * @param outputURI        Output URI or file path (directory containing CSV
+  *                         file)
+  * @param sparkMasterName  Spark master name, e.g. "local[1]"
+  * @param params           Additional parameters, currently:
+  *                         params(0): The DPLA MAP field to analyze
+  */
+class PropertyValueReport (
+                            val inputURI: String,
+                            val outputURI: String,
+                            val sparkMasterName: String,
+                            val params: Array[String]
+                          ) extends Report {
+
+  /*
+   * We set instance fields from constructor arguments, and override accessor
+   * methods in order to make the class more amenable to unit testing.
+   * However ...
+   * FIXME: It's not clear how to write unit tests that involve Dataframes
+   * and Datasets, so there are no such tests for this class.
+   *
+   */
+  override val sparkAppName: String = "PropertyValueReport"
+  override def getInputURI: String = inputURI
+  override def getOutputURI: String = outputURI
+  override def getSparkMasterName: String = sparkMasterName
+  override def getParams: Option[Array[String]] = {
+    if (params.nonEmpty) Some(params) else None
+  }
+
+
+  /**
+    * Process the incoming dataset.
+    *
+    * @see          Report.process()
+    *
+    * @param ds     Dataset of DplaMapData (mapped or enriched records)
+    * @param spark  The Spark session, which contains encoding / parsing info.
+    * @return       DataFrame, typically of Row[value: String, count: Int]
+    */
+  override def process(ds: Dataset[DplaMapData],
+                       spark: SparkSession
+                      ): DataFrame = {
+
+    import spark.implicits._
+
+    val token: String = getParams match {
+      case Some(p) => p.head
+      case _ => throw new RuntimeException(s"No field specified")
+    }
+
+    token match {
+        /*
+         * FIXME: "java.lang.UnsupportedOperationException: No Encoder found
+         * for java.net.URI" for sourceResource.language, subject, and other
+         * fields of classes that have URIs, even if you're not evaluating
+         * one of the URI fields in that dpla.ingestion3.model case class.
+         */
+      case "sourceResource.format" =>
+        ds.map(dplaMapData => dplaMapData.sourceResource.format)
+          .flatMap(x => x)
+          .groupBy("value")
+          .count
+      case "sourceResource.rights" =>
+        ds.map(dplaMapData => dplaMapData.sourceResource.rights)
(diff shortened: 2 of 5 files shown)


src/main/scala/dpla/ingestion3/ReporterMain.scala
```diff
@@ -0,0 +1,95 @@
...
+        Some(new PropertyValueReport(
+          inputURI, outputURI, sparkMasterName, reportParams
+        ))
+      case _ => None
+    }
+  }
+
+  def main(): Unit = {
+    val report = getReport(token)
+    if (report.isDefined) {
```
moltude: Is this clearer with pattern matching?
```
val reportResult = getReport(token) match { 
  case Some(report) => report.run() // Will return a Try[]
  case _ => Failure("error message")  // Return Failure w/o trying b/c $token not matched
}
// log errors
reportResult match {
  case Failure(ex) => logger.error(ex.toString)
  case _ => Unit
}
```

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
```diff
@@ -0,0 +1,108 @@
...
+                      ): DataFrame = {
+
+    import spark.implicits._
+
+    val params = getParams
+    if (params.nonEmpty) {
+
+      getParams.head match {
+          /*
+           * FIXME: "java.lang.UnsupportedOperationException: No Encoder found
```
moltude: IIRC this can be addressed by adding 
`.set("spark.serializer", "org.apache.spark.serializer.KryoSerializer")` when setting up SparkConf

and then adding 
`implicit val dplaMapDataEncoder = org.apache.spark.sql.Encoders.kryo[DplaMapData]` 
below the sparkConf

moltude: hmm, I see that you are doing exactly that below so scratch the above comment. I may be conflating issues.

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
```diff
@@ -0,0 +1,108 @@
...
+  /*
+   * We set instance fields from constructor arguments, and override accessor
+   * methods in order to make the class more amenable to unit testing.
+   * However ...
+   * FIXME: It's not clear how to write unit tests that involve Dataframes
+   * and Datasets, so there are no such tests for this class.
+   *
+   */
+  override val sparkAppName: String = "PropertyValueReport"
+  override def getInputURI: String = inputURI
```
moltude: Should these be Option(String) Option(Array[String]) to make accessing these properties (esp. params) more scala-like?

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
```diff
@@ -0,0 +1,108 @@
...
+    * @return       DataFrame, typically of Row[value: String, count: Int]
+    */
+  override def process(ds: Dataset[DplaMapData],
+                       spark: SparkSession
+                      ): DataFrame = {
+
+    import spark.implicits._
+
+    val params = getParams
+    if (params.nonEmpty) {
```
moltude: If params changed to Option(Array[String]) this could become clearer. Alternatively, you could call `headOption` instead of `head` to address the Codacy complaint

src/main/scala/dpla/ingestion3/ReporterMain.scala
markbreedlove: Yes, that's better, let me try that. Codacy complains about the `report.get()`, too

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
markbreedlove: I think I considered that, but wanted them to fail hard in the case of no input file, etc.

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
markbreedlove: Yes, I'll do the syntax you described above where you do `case Some(x) => x.something`

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
```diff
@@ -0,0 +1,108 @@
...
+           * FIXME: "java.lang.UnsupportedOperationException: No Encoder found
+           * for java.net.URI" for sourceResource.language, subject, and other
+           * fields of classes that have URIs, even if you're not evaluating
+           * one of the URI fields in that dpla.ingestion3.model case class.
+           */
+        case "sourceResource.format" =>
+          ds.map(dplaMapData => dplaMapData.sourceResource.format)
+            .flatMap(x => x)
+            .groupBy("value")
+            .count
```
AudreyAltman: Instead of doing a `map` and `flatMap` here, could you do something like `ds.select("sourceResource.format").where("format is not null").groupBy("value").count`?  I haven't played with a DplaMapData DataSet yet, but it seems like you shouldn't have to run `map` over a `DataSet` for this type of transformation.

src/main/scala/dpla/ingestion3/reports/Report.scala
```diff
@@ -0,0 +1,91 @@
...
+      SparkSession.builder().config(sparkConf).getOrCreate()
+    val sc = spark.sparkContext
+
+    val input: Dataset[DplaMapData] =
+      spark.read.avro(getInputURI).as[DplaMapData]
+    val output: DataFrame = process(input, spark)
+
+    Utils.deleteRecursively(new File(getOutputURI))
+    output
+      .repartition(1)  // Otherwise multiple CSV files
```
AudreyAltman: I think if you use `coalesce` instead of `repartition` here it will be a bit more efficient b/c it will avoid a full shuffle of the data.

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
```diff
@@ -0,0 +1,108 @@
...
+         * FIXME: "java.lang.UnsupportedOperationException: No Encoder found
+         * for java.net.URI" for sourceResource.language, subject, and other
+         * fields of classes that have URIs, even if you're not evaluating
+         * one of the URI fields in that dpla.ingestion3.model case class.
+         */
+      case "sourceResource.format" =>
+        ds.map(dplaMapData => dplaMapData.sourceResource.format)
+          .flatMap(x => x)
+          .groupBy("value")
+          .count
```
AudreyAltman: Instead of doing a `map` and `flatMap` here, could you do something like `ds.select("sourceResource.format").where("format is not null").groupBy("value").count`?  I haven't played with a DplaMapData DataSet yet, but it seems like you shouldn't have to run `map` over a `DataSet` for this type of transformation.

src/main/scala/dpla/ingestion3/reports/Report.scala
mdellabitta: 👍

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
mdellabitta: I'm not sure if those DataFrame operations work on DataSets or not (I'm new to this). But I also don't think the `select`/`where` syntax is really much clearer than what's written, and it's also not checked by the compiler in the same way, so I'd actually argue for using `map`. But`.flatten` means the same thing as `.flatMap(x => x)` and is tidier. Also the function in the map call can be shortened to `map(_.sourceResource.format)` I believe.

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
```diff
@@ -0,0 +1,108 @@
...
+    * @param spark  The Spark session, which contains encoding / parsing info.
+    * @return       DataFrame, typically of Row[value: String, count: Int]
+    */
+  override def process(ds: Dataset[DplaMapData],
+                       spark: SparkSession
+                      ): DataFrame = {
+
+    import spark.implicits._
+
+    val token: String = getParams match {
```
mdellabitta: I think you could replace this with `getParams.headOption.getOrElse(...)`

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
```diff
@@ -0,0 +1,108 @@
...
+      case Some(p) => p.head
+      case _ => throw new RuntimeException(s"No field specified")
+    }
+
+    token match {
+        /*
+         * FIXME: "java.lang.UnsupportedOperationException: No Encoder found
```
mdellabitta: Yeah, this is part of the Avro serialization problem. I need to fix it.

src/main/scala/dpla/ingestion3/reports/Report.scala
```diff
@@ -0,0 +1,91 @@
...
+  * Report:  trait with common fields and methods for QA reports.
+  *
+  * Report handles the common work of loading the input Avro file and writing
+  * the result of the report's computation out to a CSV file.
+  *
+  * It stipulates an abstract method process() that must be overridden by
+  * extending classes to do the actual computation.
+  *
+  * The implementation of Report as a Trait was supposed to make it easier to
+  * unit test, but, it's been difficult to envision how to mock up the Spark
```
mdellabitta: I don't think we should mock Spark.

src/test/scala/dpla/ingestion3/reports/PropertyValueReportTest.scala
```diff
@@ -0,0 +1,17 @@
...
+import org.scalamock.scalatest.MockFactory
+import org.apache.spark.sql.{DataFrame, Row, SparkSession, Dataset}
+import dpla.ingestion3.model._
+import dpla.ingestion3.reports._
+
+
+class PropertyValueReportTest extends FlatSpec with MockFactory{
+
+  "PropertyValueReport.process" should "x" in {
+    // FIXME: How do you mock Dataframes and DataSets??
```
mdellabitta: Why not just run spark in the test?

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
markbreedlove: All I recall at the moment is I check a variety of syntaxes based on many incompatible examples on the web and just landed on something that worked. Could revisit this later.

src/main/scala/dpla/ingestion3/reports/Report.scala
markbreedlove: i'll change that. seems like it could be a performance issue and therefore should be fixed before we run it on anything substantial.

src/main/scala/dpla/ingestion3/ReporterMain.scala
```diff
@@ -0,0 +1,98 @@
...
+  def main(): Unit = {
+    val reportResult: Try[Unit] = getReport(token) match {
+      case Some(report) => report.run()
+      case None => Failure(
+        new RuntimeException(s"Report type $token is unknown")
+      )
+    }
+    reportResult match {
+      case Failure(ex) =>
+        logger.error(ex.toString)
```
markbreedlove: I agree, will change

src/test/scala/dpla/ingestion3/reports/PropertyValueReportTest.scala
markbreedlove: We want unit tests to run quickly and be runnable with Travis.

src/main/scala/dpla/ingestion3/reports/Report.scala
mdellabitta: We'd have to experiment to know that's true. In general, `coalesce` can mean that the data is manipulated without any cross-node traffic, but that's not true of `coalesce(1)`. Not a hard change, however.

src/test/scala/dpla/ingestion3/reports/PropertyValueReportTest.scala
mdellabitta: I don't think we'd not get that with embedded Spark. If we don't, however, I'd argue for moving this to an integration test over mocking Spark.

src/main/scala/dpla/ingestion3/reports/PropertyValueReport.scala
AudreyAltman: This is an academic comment, since the PR has already been merged, but for future reference I wanted to leave a clarification here (since I left for vacation before the thread closed).  My understanding is that lambda functions can be less efficient than DataFrame operations when using DataSets.  I believe that the Catalyst Optimizer works better with DataFrame operations.  So in the above example, I think the lambda function forces Spark to encode the entire `dplaMapData` as a JVM object, whereas the `where` filter allows it to pick out the `sourceResource.format` field and only encode that.  I'm definitely in newbie territory here, but have gotten the general impression that `map` functions are best-suited for RDDs b/c they are opaque to the Catalyst Optimizer.

mdellabitta: The DplaMapData is always a JVM object because there's no encoder for it given what Spark makes available right now, unless we use Kyro, but then it's just `Array[Byte]` in memory, and there's no columnar serialization. So there's no opportunity for Catalyst to do anything fancy anyway w/r/t predicate pushdown right now. Touching a row means deserializing the whole thing.

The reason we're dealing with something that doesn't have an Encoder is to get those compile time checks on `Option` fields and those alias types for cardinality so forth, so we've pretty consciously gone down this path to get these sorts of type checking benefits vs. the nullable, constrained type fields of a Row.

Because we deal in the land of Strings and free text instead of 16-bit floats (ML data), I doubt the performance benefits we'd get from writing our domain logic against the normal Row/StructField schema stuff would be worth losing those compile checks, but I could be wrong. We'll get an opportunity to check that in theory when I finish the Avro story, because it turns out the best way to accomplish that is to desugar the DplaMapData model to a Row/StructField schema record and run that through the Avro serialization addon.

AudreyAltman: Thanks @mdellabitta I had forgotten about the encoding issues with the `Option` fields.  That all makes sense.

PR #57, opened 2017-08-07
Update documentation to bring in line w/wiki

moltude:
Updates README.md with the documentation @AudreyAltman created for https://digitalpubliclibraryofamerica.atlassian.net/wiki/spaces/TECH/pages/87658172/Spark+OAI+Harvester

The invocations described on the README and Wiki are slightly different. I found SBT to be the easiest way to invoke the harvester so that is what I included here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,67 +1,85 @@
 # DPLA Ingestion 3
 
-##Overview
-This project is an ETL system for cultural heritage metadata. 
+This project is an ETL system for cultural heritage metadata. The system has three primary components:
 
 ### Harvesters
 * [OAI-PMH](#oai-pmh)
-* ResourceSync
-* API
-* File
-
+* [ResourceSync]() TBD
+* [API]() TBD
+* [File]() TBD
+* [Primo]() TBD
 ### Mappers
-* ID minting  
-
+* [TBD]() 
 ### Enrichments
-* String enrichments
-* Linked data enrichments
+* [String enrichments]()
+* [Linked data enrichments]()
 
 
 
-## OAI-PMH
-The OAI harvester takes its parameters from a config file that is passed in as a command line argument. 
+## OAI-PMH Harvester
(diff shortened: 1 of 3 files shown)


moltude:
@markbreedlove FYI -- I forgot to get the link to the blog post. Should add that before merging

src/main/resources/profiles/digital-commonwealth.conf
```diff
@@ -11,6 +11,5 @@ endpoint = "http://fedora.digitalcommonwealth.org/oaiprovider/"
 metadataPrefix = "mods"
 outputDir = "" # TODO This should be an S3:// bucket that i3 has write access to
 provider = "Digital Commonwealth"
-type = "oai" # Which harvester to invoke
```
markbreedlove: Since the scope of the PR is to update just the README, can you comment on the removal of this line and the one in `oaiSample.conf`?

README.md
```diff
@@ -1,68 +1,85 @@
...
-provider = "dpla hub"
-# One of these options must be set: blacklist, whitelist or harvestAllSets
-blacklist = "badSet1,badSet2"
-# whitelist = "" 
-# harvestAllSets = true
+# oai-sample.conf
 verb = "ListRecords"
+endpoint = "http://fedora.sample.org/oaiprovider/"
+metadataPrefix = "mods"
+outputDir = "/path/to/somewhere"
```
markbreedlove: `outputDir` is not in the table above. Could this also be an s3 URL?

src/main/resources/profiles/digital-commonwealth.conf
markbreedlove: It's clear that we only have one harvester so far, OAI-PMH, but won't there be others at some point, such that the `type` option still has a place?

moltude: I removed it because we are not actively using it and didn't want to presume that this was going to be the implementation before we actually did it.

README.md
moltude: done

AudreyAltman:
This needs a rebase before merging.

PR #58, opened 2017-08-09
Fix error handling

moltude:
Rewrite error handling to quit on errors that are unparsable (XML parsing, OAI error, HTTP error). Previously, these errors were being swallowed and leading to an infinite loop where the first page was being requested over and over despite being an error.

No longer embeds entire response page in every single extracted item. This was causing OOM exceptions.

### DIFF
--- a/src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
@@ -40,7 +40,9 @@ object OaiResponseProcessor {
           case Some(node) =>
             val id = getRecordIdentifier(node)
             val setIds = getSetIdsFromRecord(node)
-            Some(OaiRecord(id, node.toString, setIds, source))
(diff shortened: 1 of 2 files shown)


src/main/scala/dpla/ingestion3/harvesters/oai/OaiResponseProcessor.scala
```diff
@@ -40,7 +40,9 @@ object OaiResponseProcessor {
           case Some(node) =>
             val id = getRecordIdentifier(node)
             val setIds = getSetIdsFromRecord(node)
-            Some(OaiRecord(id, node.toString, setIds, source))
+            // Drop the entire page text from source when returning OaiRecord. Embedding the entire
+            // page text was triggering Out of Memory exceptions even on small harvests (200,000 records)
+            Some(OaiRecord(id, node.toString, setIds, OaiSource(source.queryParams, url=source.url)))
```
mdellabitta: Prolly a clearer way of saying "remove the text" is `source.copy(text = None)` rather than `OaiSource(source.queryParams, url=source.url)`

moltude:
@mdellabitta fixed.

PR #60, opened 2017-08-10
PA Extractor

moltude:
Adds the Pennsylvania extractor.  There are a lot of TODOs and follow-up points here but the core requirement of DT-1309 is addressed in this PR.

The Extractor class was rewritten so that #build() returns a DplaMap class which is inclusive of DplaMapData (success) and DplaMapError (failed mapping).

Additionally, this work may require a revision after reviewing changes to `DplaMapData`.

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/PaExtractor.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/PaExtractor.scala
@@ -3,62 +3,82 @@ package dpla.ingestion3.mappers.providers
 import java.net.URI
 
 import dpla.ingestion3.mappers.xml.XmlExtractionUtils
-import dpla.ingestion3.model.DplaMapData.ZeroToOne
+import dpla.ingestion3.model.DplaMapData.{ExactlyOne, ZeroToOne}
 import dpla.ingestion3.model._
 
+import scala.util.{Failure, Success, Try}
 import scala.xml._
 
 class PaExtractor(rawData: String) extends Extractor with XmlExtractionUtils {
 
-  implicit val xml: NodeSeq = XML.load(rawData)
+  implicit val xml: NodeSeq = XML.loadString(rawData)
 
   def agent = EdmAgent(
     name = Some("Pennsylvania Digital Collections Project"),
     uri = Some(new URI("http://dp.la/api/contributor/pa"))
   )
 
-  // Get the last occurrence of the identifier property
+  // Get the last occurrence of the identifier property, there
+  // must be at least three dc:identifier properties for there
+  // to be a thumbnail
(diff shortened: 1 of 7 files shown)


src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -95,10 +106,15 @@ case class OreAggregation(
                            originalRecord: ExactlyOne[String], //map v4 specifies this as a ref, but that's LDP maybe?
                            hasView: ZeroToMany[EdmWebResource] = Seq(),
                            intermediateProvider: ZeroToOne[EdmAgent] = None,
-                           `object`: ZeroToOne[EdmWebResource] = None, //full item page
-                           preview: ZeroToOne[EdmWebResource] = None, ///thumbnail
+                           `object`: ZeroToOne[EdmWebResource] = None, // full size image (should *not* be used)
```
mdellabitta: I think the take away is we do want to use this if available in the raw data, but generally thus far it is not.

src/main/scala/dpla/ingestion3/mappers/providers/PaExtractor.scala
```diff
@@ -3,57 +3,96 @@ package dpla.ingestion3.mappers.providers
...
+          originalRecord = rawData,
+          provider = agent,
+          preview = thumbnail
+          // TODO what about `object` in this context? Assume no implementation
+        )
       )
-    )
+    } match {
+      case Success(s) => s
+        // TODO Is there a way to write this so that we don't have to write this match block into every Extractor?
```
mdellabitta: yeah, we could have an implicit function in scope that turns a Try[DplaMap] into either a DplaMapData or a DplaMapError, moving everything within match {} to that.

moltude: @mdellabitta Took a crack at implementing that in `Extractor`. Tested against PA mapping with same results.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
moltude: I'm still not sure I can keep this straight in my head. I don't think there is a ticket to do the modeling review (next sprint?). I'll do that tomorrow morning and tag this for that review.

moltude:
If @mdellabitta is signing off I think I can take off `DO NOT MERGE` ;). Will still hold for at least one more review.

mdellabitta:
I guess one thing is that Codacy warning. I'd use `.getOrElse("ID NOT FOUND")` or something instead, so you don't accidentally mask the exception you wanted to throw with a NullPointerException or whatever instead.

AudreyAltman:
This needs to be rebased before merging.

PR #61, opened 2017-08-11
Add PropertyValueReport

moltude:
This report takes a property and builds a Dataset that contains the values of the
property, the provider's local uri and the DPLA uri.

Refactors PropertyValueReport class name to PropertyDistinctValueReport because it is
only returning the unique values. Also to distinguish it from the PropertyValueReport
where all values for a given property are returned. This name change was suggested as
a part of DT-1498 (see Gretchen's comment)

### DIFF
--- a/src/main/scala/dpla/ingestion3/ReporterMain.scala
+++ b/src/main/scala/dpla/ingestion3/ReporterMain.scala
@@ -26,10 +26,11 @@ import util.{Try, Failure}
   *
   * @see ReporterTest
   *
-  * @param token            The report name token
+  *
   * @param inputURI         Input URI or file path
   * @param outputURI        Output URI or file path
   * @param sparkMasterName  Spark master name, e.g. "local[1]"
+  * @param token            The report name token
   * @param reportParams     Additional parameters particular to the report
   */
 class Reporter (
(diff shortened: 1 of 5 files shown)


src/main/scala/dpla/ingestion3/ReporterMain.scala
```diff
@@ -26,10 +26,11 @@ import util.{Try, Failure}
   *
   * @see ReporterTest
   *
-  * @param token            The report name token
+  *
   * @param inputURI         Input URI or file path
   * @param outputURI        Output URI or file path
   * @param sparkMasterName  Spark master name, e.g. "local[1]"
+  * @param token            The report name token
```
moltude: This reordering was just to bring it in line with how it expects the parameters to be ordered.

src/main/scala/dpla/ingestion3/reports/PropertyDistinctValueReport.scala
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    token match {
+        /*
+         * FIXME: "java.lang.UnsupportedOperationException: No Encoder found
+         * for java.net.URI" for sourceResource.language, subject, and other
+         * fields of classes that have URIs, even if you're not evaluating
+         * one of the URI fields in that dpla.ingestion3.model case class.
+         */
+      case "sourceResource.format" =>
```
mdellabitta: Was thinking of suggesting some DRY fix here, but it's quite probable that the actual code pipeline here will vary based on the cardinality of the field being extracted. At some point, we might want to extract a pattern or two, however.

PR #62, opened 2017-08-16
Fix for Avro serialization

mdellabitta:
This PR contains two functions, one to convert a DplaMapModel to a Row, and one to convert a Row into a DplaMapModel. These can be used at the point when we need to write to Avro and  to subsequently read from those Avro files. Additionally, we can in theory use these to serialize to any other format Spark supports (e.g., Parquet), and we can work with the data in desugared, primitive format in SparkSQL if needed.

### DIFF
--- a/src/main/scala/dpla/ingestion3/model/DplaMapData.scala
+++ b/src/main/scala/dpla/ingestion3/model/DplaMapData.scala
@@ -16,7 +16,6 @@ object DplaMapData {
   type ZeroToOne[T] = Option[T]
   type ExactlyOne[T] = T
   type LiteralOrUri = Either[String,URI]
-  type LiteralOrSkos = Either[String,SkosConcept]
 }
 
 sealed trait DplaMap
@@ -150,8 +149,7 @@ case class DplaPlace(
                      county: ZeroToOne[String] = None,
                      state: ZeroToOne[String] = None,
                      country: ZeroToOne[String] = None,
-                     // Region or equivalent is not returned by Twofishes
-                     // region: ZeroToOne[String] = None,
+                     region: ZeroToOne[String] = None,
                      coordinates: ZeroToOne[String] = None
                     )
 
--- a/src/main/scala/dpla/ingestion3/model/RowConverter.scala
+++ b/src/main/scala/dpla/ingestion3/model/RowConverter.scala
@@ -1,0 +1,118 @@
+package dpla.ingestion3.model
+
+import dpla.ingestion3.model.DplaMapData.LiteralOrUri
+import org.apache.spark.sql.Row
+import org.apache.spark.sql.catalyst.expressions.GenericRowWithSchema
+import org.apache.spark.sql.types.{ArrayType, StringType, StructField, StructType}
+
+/**
+  * Responsible for desugaring a DplaMapModel and converting it to a Spark-native Row-based structure.
+  *
+  * PLEASE NOTE: The *ordering* of the parameters to each call to the Row() constructor is significant, and
+  * cannot be changed without a concurrent effort to rewrite all the data in the master dataset, and a matching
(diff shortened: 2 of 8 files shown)


mdellabitta:
Codacy's complaining about null a lot, but I don't really have a great way of dealing with that becasuse null is how you say things aren't present in a Row.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -139,8 +138,7 @@ case class DplaPlace(
                      county: ZeroToOne[String] = None,
                      state: ZeroToOne[String] = None,
                      country: ZeroToOne[String] = None,
-                     // Region or equivalent is not returned by Twofishes
-                     // region: ZeroToOne[String] = None,
+                     region: ZeroToOne[String] = None,
```
moltude: Curious -- What was the reason for uncommenting this line for this specific PR?

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -15,7 +15,6 @@ object DplaMapData {
   type ZeroToOne[T] = Option[T]
   type ExactlyOne[T] = T
   type LiteralOrUri = Either[String,URI]
-  type LiteralOrSkos = Either[String,SkosConcept]
```
moltude: Assuming this was removed b/c it wasn't being used anywhere in the Model?

src/main/scala/dpla/ingestion3/model/RowConverter.scala
```diff
@@ -0,0 +1,109 @@
...
+    oa.hasView.map(edmWebResource),
+    oa.intermediateProvider.map(edmAgent).orNull,
+    oa.`object`.map(edmWebResource).orNull,
+    oa.preview.map(edmWebResource).orNull,
+    edmAgent(oa.provider),
+    oa.edmRights.map(_.toString).orNull
+  )
+
+  private[model] def dplaSourceResource(sr: DplaSourceResource): Row = Row(
+    sr.alternateTitle,
```
moltude: The order of fields really matters here, right? This ordering here determines the index value when converting from Row to Map? If this is the case I think a comment explaining that would be very helpful esp. when we need to update the model to add a field and maintain alphabetical order of field names.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
mdellabitta: That is true, but also the way you'd make a SkosConcept with only a string would be to assign it to the concept field like `SkosConcept(concept = "a concept")`, not make a union type. Just don't want people to get confused.

src/main/scala/dpla/ingestion3/model/RowConverter.scala
mdellabitta: Yes, that's a great point.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
mdellabitta: I'm laying this out in a order sensitive `Row` as you mention below, so I had to basically bake in a somewhat final representation of everything, and needed to make some choices.

I don't think there was a good reason to remove this field previously... it's optional, it's in the MAP spec, and could easily be satisfied with a lookup dictionary of country -> region (i.e., we don't need Twofishes to support it for it to have a reason to exist and a way for it to be populated). So I put it back in so it could sit in the natural order of small to big fields and we wouldn't have to do any complicated field shuffling to get it back.

src/main/scala/dpla/ingestion3/model/RowConverter.scala
mdellabitta: OK I pushed something up to address this point.

PR #64, opened 2017-08-24
FIX: Include empty/missing values to all reports and add support for all fields

moltude:
This PR addresses both DT-1546 and DT-1545. 

All sourceResource fields were added to the PropertyValueReport and PropertyDistinctValueReport. Case classes were also added to both report classes to tighten up the code. Adds an extraction method for the report values. This method is also used to extract the correct field from all the context case classes.

### DIFF
--- a/src/main/scala/dpla/ingestion3/reports/PropertyDistinctValueReport.scala
+++ b/src/main/scala/dpla/ingestion3/reports/PropertyDistinctValueReport.scala
@@ -1,8 +1,11 @@
 package dpla.ingestion3.reports
 
-import org.apache.spark.sql.{DataFrame, Row, SparkSession, Dataset}
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
 import dpla.ingestion3.model._
+import org.apache.spark.sql.functions.{col, explode}
 
+
+case class PropertyDistinctValueRpt(value: Seq[String])
 /**
   * Property Distinct Value QA report.  Takes one field and gives a count of unique
   * values.
@@ -38,7 +41,7 @@ class PropertyDistinctValueReport(
                             val outputURI: String,
                             val sparkMasterName: String,
                             val params: Array[String]
-                          ) extends Report {
+                          ) extends Report with Serializable {
 
   /*
    * We set instance fields from constructor arguments, and override accessor
@@ -55,7 +58,6 @@ class PropertyDistinctValueReport(
   override def getParams: Option[Array[String]] = {
     if (params.nonEmpty) Some(params) else None
   }
-
 
   /**
     * Process the incoming dataset.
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/reports/PropertyDistinctValueReport.scala
```diff
@@ -77,32 +79,82 @@ class PropertyDistinctValueReport(
...
+    * Explodes the value column so that multiple values
+    * occur on separate rows. Then the groupBy and count
+    * operations are performed on the data
+    *
+    * @param rptDataset
+    * @param spark
+    * @param token
+    * @return
+    */
+  def makeCsvTable(rptDataset: Dataset[PropertyDistinctValueRpt],
```
AudreyAltman: This may a bit nit-picky, but would you consider calling this method just `makeTable` since it's not actually writing to a CSV file?  I was a bit confused at first.

src/main/scala/dpla/ingestion3/reports/PropertyDistinctValueReport.scala
```diff
@@ -77,32 +79,82 @@ class PropertyDistinctValueReport(
...
+                   spark: SparkSession,
+                   token: String): DataFrame = {
+    val sqlContext = spark.sqlContext
+    // Periods are not allowed in column names...
```
AudreyAltman: Thanks for documenting this!

src/main/scala/dpla/ingestion3/reports/Report.scala
```diff
@@ -88,4 +89,38 @@ trait Report {
...
+        case _: EdmAgent =>
+          t.map(_.asInstanceOf[EdmAgent].name.getOrElse("__MISSING EdmAgent.name__"))
+        case _: EdmTimeSpan =>
+          t.map(_.asInstanceOf[EdmTimeSpan].originalSourceDate.getOrElse("__MISSING EdmTimeSpan.originalSourceDate__"))
+        case _: SkosConcept =>
+          t.map(_.asInstanceOf[SkosConcept].providedLabel.getOrElse("__MISSING SkosConcept.providedLabel__"))
+        case None => Seq("__MISSING__")
+        case _ => t.map(_.toString)
+      }
+    }
```
AudreyAltman: This looks useful -- like something we could re-use in other reports.

src/main/scala/dpla/ingestion3/reports/PropertyDistinctValueReport.scala
```diff
@@ -77,32 +79,82 @@ class PropertyDistinctValueReport(
...
+    val colName = token.replace(".", "_")
+    rptDataset.createOrReplaceTempView("tmpPropValRpt")
+
+    sqlContext.sql("""SELECT * FROM tmpPropValRpt""")
+      .withColumn(colName, explode(col("value")))
```
AudreyAltman: Exploding `value`!  Clever.

moltude:
@AudreyAltman makeCsvTable name refactored.

PR #65, opened 2017-08-28
thumbnail report

AudreyAltman:
This adds thumbnail QA reports.  It addresses [ticket 1501](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1501).  It has been tested locally with PA data.

### DIFF
--- a/src/main/scala/dpla/ingestion3/reports/ThumbnailReport.scala
+++ b/src/main/scala/dpla/ingestion3/reports/ThumbnailReport.scala
@@ -1,0 +1,202 @@
+package dpla.ingestion3.reports
+
+import dpla.ingestion3.model.DplaMapData
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
+import scala.util.{Try, Failure, Success}
+import javax.imageio.ImageIO
+import java.net.URI
+
+/**
+  * Produces QA reports related to thumbnails (i.e. oreAggregation.preview).
+  *
+  * The "missing" report (i.e. params = "missing") returns all images that are
+  * missing thumbnails.
+  * The resulting Dataframe has the following columns:
+  *   - localUri: String, the provider URI for the item
+  *   - dplaUri: String, DPLA's URI for the item
+  *
+  * The "dimensions" report (i.e. params = "dimensions") returns the height and
+  * width of each image, plus any errors encountered when attempting to call the
+  * URI endpoint.  Please note that this report takes a long time to run, as it
+  * has to make an HTTP request for every thumbnail in the data sample.
+  * The resulting Dataframe has the following columns:
+  *    - localUri: String, the provider URI for the item
+  *    - dplaUri: String, DPLA's URI for the item
+  *    - preview: String, thumbnail URI
+  *    - height: Int, height in pixels of the image
+  *    - width: Int, width in pixels of the image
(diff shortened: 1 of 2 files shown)


src/main/scala/dpla/ingestion3/reports/ThumbnailReport.scala
```diff
@@ -0,0 +1,212 @@
...
+    * Warning: This process takes a long time to run.
+    */
+  def dimensionsReport(ds: Dataset[DplaMapData], spark: SparkSession): DataFrame = {
+    import spark.implicits._
+
+    val thumbnailData: Dataset[ThumbnailDimensions] = ds.map(dplaMapData => {
+
+      val uri: Option[URI] = previewUri(dplaMapData)
+
+      val preview: Option[String] = uri match {
```
mdellabitta: I think you can just do `uri.map(_.toString)` here.

src/main/scala/dpla/ingestion3/reports/ThumbnailReport.scala
```diff
@@ -0,0 +1,212 @@
...
+        case Some(uri) => Some(uri.toString)
+        case None => None
+      }
+
+      val dimensions: Option[Dimensions] = uri match {
```
mdellabitta: Similarly, you can do `uri.map(getDimensions)`

src/main/scala/dpla/ingestion3/reports/ThumbnailReport.scala
```diff
@@ -0,0 +1,212 @@
...
+    * @param uri URI for the thumbnail
+    * @return Dimensions case class containing the height and width, or an error
+    *         message if an error occurs.
+    */
+  def getDimensions(uri: URI): Dimensions = {
+    bufferedImg(uri) match {
+      case Success(bimg) =>
+        // Checking for null here creates a codacy error, but since BufferedImage
+        // is a Java class, it can legitimately be null.
+        if (bimg == null) Dimensions(error = Some("Buffered image is null."))
```
mdellabitta: Stupid Java! I'm thinking of writing a function somewhere that just returns `null` so we can use it instead of `null` everywhere else and have only one warning about it.

src/main/scala/dpla/ingestion3/reports/ThumbnailReport.scala
```diff
@@ -0,0 +1,212 @@
...
+    */
+  def dplaUri(dplaMapData: DplaMapData): String =
+    dplaMapData.oreAggregation.uri.toString
+
+  /**
+    * @param dplaMapData
+    * @return Option[URI], the thumbnail URI
+    */
+  def previewUri(dplaMapData: DplaMapData): Option[URI] =
+    dplaMapData.oreAggregation.preview match {
```
mdellabitta: you can use that .map trick here, too

src/main/scala/dpla/ingestion3/reports/ThumbnailReport.scala
AudreyAltman: Clever idea!

AudreyAltman:
Thanks, @mdellabitta, I made those style fixes you suggested.

mdellabitta:
👍 :shipit:

PR #66, opened 2017-08-29
fix oai harvester invocation

AudreyAltman:
This fixes the OAI harvester invocation (at least for me).  

The OAI harvester invocation was not working on my local machine.  Using `-Dconfig` in the SBT command line did not seem to be successfully setting the system properties, so when `ConfigFactory.load()` was called, my `.conf` file was not being read in.  I don't know _why_ this was happening, but this PR fixes it.  The same bug was reported by one of our users in the tech channel.  Scott was _not_ having trouble with the previous implementation.

Instead of using `-Dconfig` in the command line, this new PR reads in a file locally or from a URL, and sets the system property in `OaiHarvesterConf`.

The user does not need to specify whether a file is local or a URL; the application checks this.  The upside of having the application check for a URL is that it's a simpler command-line interface.  The downside is that there is no way to specify reading a file from a classpath, as there was before.  I'm not sure if we need to account for this use case - if so, we could add a command line arg specifying "file", "url", or "classpath".

I have tested this using a local `.conf` file, and one in S3.  It would be good to get at least one more test from someone on the team (and maybe the person from the tech channel).

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,24 +61,21 @@ To use SBT you need to specify the path to the config file you just created when
 
 ```
 # Example when the path is stored locally. This should address 95% of all use cases
--Dconfig.file=/loca/path/to/config.conf
+/local/path/to/config.conf
 
-# Exmple when the profile is stored remotely
--Dconfig.url=https://s3.amazonaws.com/dpla-i3-oai-profiles/sample_provider.conf
-
-# Example when the config file is on the project classpath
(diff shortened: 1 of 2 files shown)


AudreyAltman:
FYI here's a link to the error reported in the slack channel: https://dplatech.slack.com/archives/C51NH2CF3/p1503330679000386

AudreyAltman:
@mdellabitta yes that is correct.

PR #70, opened 2017-08-31
Xml Shredder

AudreyAltman:
This adds a pre-mapping report to shred harvested XML docs.  It addresses [ticket DT-1488](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1488).

The report includes both `xmlns` data and prefixes for labels (ie. a label will be `record.about.oaiProvenance:provenance` instead of `record.about.provenance`.  This can be easily changed if it's too overwhelming for QA workflows.

I tried to write tests for `XmlShredder` but was running into some serialization issues within the test environment that I did not know how to solve.  I did test this locally with a sample of harvested data.

### DIFF
--- a/src/main/scala/dpla/ingestion3/PreMappingReporterMain.scala
+++ b/src/main/scala/dpla/ingestion3/PreMappingReporterMain.scala
@@ -1,0 +1,89 @@
+package dpla.ingestion3
+
+import dpla.ingestion3.premappingreports._
+import org.apache.log4j.{LogManager, Logger}
+import scala.util.{Failure, Try}
+
+/**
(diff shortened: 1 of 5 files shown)


AudreyAltman:
@mdellabitta I've made the changes we discussed, including:
1. Include the doc id in the data output.
2. Use `@attribute`, `@xmlns`, and `text()` notation
3. Change the expected input to the established harvested data model as opposed to the more complex OAI data model.
4. Create a stand-alone `XmlShredder` object that is unaware of the spark context; create an `XmlShredderReport` class to hand the spark context.
4. Fix indentation.

moltude:
Ran against MA data (490,428 records) and the output file was 13G 💥

PR #71, opened 2017-09-05
fix block comment in ma conf

AudreyAltman:
I got an error when running this b/c of the block comment.  I don't know of a supported way to do block comments, so I changed it to single-line and it works for me now.


moltude:
@AudreyAltman 👍

PR #72, opened 2017-09-06
Minnesota Digital Library harvester

moltude:
Addresses DT-1505
    * Adds the MDL API harvester
    * Refactor the CdlResponse trait and corresponding case classes to ApiResponse
    * Add ApiHarvester trait that CDL and MDL both extend with common method saveOut()
    * ApiHarvester also includes a common invocation for running API harvests


src/main/scala/dpla/ingestion3/CdlHarvesterMain.scala
```diff
@@ -36,111 +34,77 @@ object CdlHarvesterMain {
...
-    val rows = conf.rows.getOrElse("100")
-    // Same as apiKey, this is validated at invocation.
-    val endpoint = conf.endpoint.getOrElse("MISSING ENDPOINT")
-
-    implicit val formats = DefaultFormats
+    val queryParams = Map(
+      "query" -> conf.query.getOrElse("*:*"),
+      "rows" -> conf.rows.getOrElse("100"),
+      "api_key" -> conf.apiKey.getOrElse("MISSING API KEY")
+    )
```
moltude: I removed endpoint from the API configs and chose to hard code it while building the query URLs in each harvester since they are provider specific.

src/main/scala/dpla/ingestion3/CdlHarvesterMain.scala
```diff
@@ -36,111 +34,77 @@ object CdlHarvesterMain {
...
     val schemaStr = new FlatFileIO().readFileAsString("/avro/OriginalRecord.avsc")
     val schema = new Schema.Parser().parse(schemaStr)
     val avroWriter = getAvroWriter(outFile, schema)
 
-    val queryParams = Map(
-      "query" -> query,
-      "rows" -> rows,
-      "api_key" -> apiKey
-    )
+    startHarvest(outFile, queryParams, avroWriter, schema)
```
moltude: Calls #startHarvest() in ApiHarvester

moltude:
@mdellabitta comments addressed

PR #77, opened 2017-09-13
json shredder

AudreyAltman:
This adds a JSON shredder, which functions like the existing XML shredder.

This has been tested locally with a random sample of ~1,000 MDL records.  It has been indexed here: http://ec2-52-87-236-197.compute-1.amazonaws.com:9200/json-shredder-test/_search?pretty=true&q=*:*

In order to get elastic search to index the records, I had to turn off the date serialization (some fields had both date and string values, and this was causing errors b/c elastics search expected all fields values to be of the same type).  Here is the invocation I used:

```
curl -XPUT 'http://ec2-52-87-236-197.compute-1.amazonaws.com:9200/json-shredder-test/' -d '{"mappings":{"_default_":{"date_detection":false}}}'
```

I made some small changes to the XML shredder for consistency, notably concatenating field names with `_` instead of `.` (the latter was causing a perplexing serialization bugs when it hit elastic search).

This addresses [ticket DT-1849](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1489).


src/main/scala/dpla/ingestion3/preMappingReports/XmlShredder.scala
```diff
@@ -168,13 +168,5 @@ object XmlShredder {
   def labelWithPrefix(node: Node): String =
     // We have to use null here b/c it can be returned by scala.xml.Node
     if(node.prefix == null) node.label
-    else s"${node.prefix}_${node.label}"
```
moltude: I'm probably missing something glaring here but why was this changed from `_` to `.`.? Everything else seems to go the other way...

AudreyAltman: Yeah, this concatenating a prefix with its label.  All the others are concatenating a parent label with a child label.  So there needs to be a different character here.

src/main/scala/dpla/ingestion3/preMappingReports/XmlShredder.scala
```diff
@@ -20,17 +20,17 @@ object XmlShredder {
     * Examples:
```
mdellabitta: Was there a find/replace in here that went awry? I thought we were using . as a path separator and _ for separating xml element names from namespace nicknames.

AudreyAltman: Yeah, I got it mixed up when I wrote the JSON shredder, and then couldn't recall which one was the original.  It's consistent now with both shredders.  Do you have strong feelings about which is path separator and which is element separator?

mdellabitta: Nope. Let's let Gretchen tell us.

PR #78, opened 2017-09-14
Add Spark properties to configuration file and logging setup

moltude:
This addresses both DT-1567, DT-1581 and DT-1578 and directly impacts DT-1566 (possibly a blocker...)

**Spark configuration and configuration generalization**
- Adds support for overriding the default Spark properties (sparkMaster, driverMemory) in the application configuration file
- These properties can be overridden within specific provider configurations
- Add generalized harvester configuration properties which will support all harvester options (API, OAI etc.)
- Updated CDL and OAI mains to use this new invocation but since @AudreyAltman is writing the single invocation this week it didn't make sense to do more than that to prove it works. 

To invoke any harvester you will need to provide three parameters:
- The harvest output location (--output)
- Path to the application configuration file (--conf)
- Abbreviated name of the provider e.g. cdl, harvard, mwdl (--name)

Sample application.conf file
```
# Default Spark configuration
spark.master="local[*]"
spark.driverMemory="2g"
spark.executorMemory="2g"

# Harvard
harvard.provider = "Harvard"
harvard.harvest.type="oai"
harvard.harvest.endpoint = "http://vcoai.lib.harvard.edu/vcoai/vc"
harvard.harvest.metadataPrefix = "mods"
harvard.harvest.setlist="dag,lap,crimes,scarlet,manuscripts,eda,cna,ihp,ward,rubbings"
harvard.harvest.verb = "ListRecords"
# Overrides the default sparkMaster value
harvard.spark.master = "local[1]"

# CDL
cdl.provider = "California Digital Library"
cdl.harvest.type="api"
cdl.harvest.rows = "500"
cdl.harvest.query = "-type_ss:dataset"
cdl.harvest.apiKey = ""
cdl.spark.master = "local[3]"
```

**Logging setup**
- Added ingestion3 logger and routes all the output to the console by default
- Uses runtime properties to create log files following the pattern <provider>-<process>-<date>.log
- Ships all Sparks logging statements to separate log files
- Cleaned up log4j.properties file

To route log messages to a custom output file called `harvard-harvest-2017-09-14.log`
```
val harvestLogger: Logger = LogManager.getLogger("ingestion3")
val appender = Utils.getFileAppender(providerName, "harvest")
harvestLogger.addAppender(appender)
harvestLogger.fatal("oh snap")
```


build.sbt
```diff
@@ -8,8 +8,16 @@ lazy val root = (project in file("."))
     Defaults.itSettings,
     libraryDependencies ++= Seq(
       "org.scalatest" %% "scalatest" % "3.0.1" % "it,test",
-      "org.apache.spark" %% "spark-core" % "2.0.1" exclude("org.scalatest", "scalatest_2.11"),
-      "org.apache.spark" %% "spark-sql" % "2.0.1" exclude("org.scalatest", "scalatest_2.11"),
+
+      /**
+        * @sw I updated spark core and spark sql from 2.0.1 to 2.1.0 to
```
mdellabitta: EMR is now on 2.2.0. I think we should keep up with it, unless there's a problem we encounter.

mdellabitta: But we can try that outside of this PR. Just saying we shouldn't be worried about having strayed from 2.0.1.

src/main/resources/log4j.properties
```diff
@@ -11,22 +10,16 @@ log4j.appender.RollingAppender.DatePattern='.'yyyy-MM-dd
...
-log4j.appender.RollingAppenderU.layout.ConversionPattern=[%p] %d %c %M - %m%n
-
+# Ingestion 3 logger
+# A FileAppender should be dynamically added to this logger
+log4j.logger.ingestion3=DEBUG, stdout
 
-# By default, everything goes to console and file
-log4j.rootLogger=ERROR, RollingAppender, stdout
+# By default, everything goes to spark.log file
+log4j.rootLogger=DEBUG, RollingAppender
```
mdellabitta: Do we want it to log at DEBUG ordinarily? That might actually slow things down.

src/main/resources/profiles/harvard.conf
```diff
@@ -1,6 +1,6 @@
 endpoint = "http://vcoai.lib.harvard.edu/vcoai/vc"
 metadataPrefix = "mods"
-outputDir = "/Users/scott/harvard_i3_output"
+outputDir = "/Users/scott/i3_harvard_harvest"
```
mdellabitta: Weren't we talking about keeping file paths outside of configuration and in command line arguments instead?

src/main/scala/dpla/ingestion3/utils/Utils.scala
```diff
@@ -100,7 +98,8 @@ object Utils {
     * @param runtime Runtime in milliseconds
     * @param recordCount Number of records output
     */
-  def printResults(runtime: Long, recordCount: Long): Unit = {
+  def logResults(runtime: Long, recordCount: Long, logger: Logger): Unit = {
+    // TODO figure out a better way to share a logger...
```
mdellabitta: We probably don't want to share loggers, but rather configure the appender and make sure different loggers use the same appender. But even then, when we're in distributed mode, that won't be possible across all the executors.

src/main/scala/dpla/ingestion3/CdlHarvesterMain.scala
```diff
@@ -32,23 +32,37 @@ object CdlHarvesterMain extends ApiHarvester {
...
-    val conf = new CdlHarvestConf(args)
-    val outFile = new File(conf.outputFile.getOrElse("out"))
-    val queryParams = Map(
-      "query" -> conf.query.getOrElse("*:*"),
-      "rows" -> conf.rows.getOrElse("100"),
-      "api_key" -> conf.apiKey.getOrElse("MISSING API KEY")
-    )
+    val cmdArgs = new HarvestCmdArgs(args)
 
+    val outputDir = cmdArgs.output.toOption match {
```
mdellabitta: Rather than using a pattern match, you can call `.map` on an option:

`cmdArgs.output.toOption.map(new File(_)).getOrElse(throw new RuntimeException("No output specified")`

src/main/resources/profiles/harvard.conf
moltude: Sorry, this was just an oversight.

src/main/scala/dpla/ingestion3/utils/Utils.scala
moltude: The other option I was considering last night was using System.properties to store what we need to dynamically name output files and I think that would allow us to just load the `ingestion3` logger from multiple classes and get the output routed correctly.

src/main/scala/dpla/ingestion3/CdlHarvesterMain.scala
moltude: 😎

src/main/resources/log4j.properties
moltude: Honestly, I really wasn't sure but I can roll it back ...

build.sbt
moltude: Thanks. To clarify, are you saying the move from 2.0.1 to 2.1.0 should be done as a separate ticket/PR?

mdellabitta: Nope, I think we should leave what you have, and do 2.2.0 in another ticket.

src/main/scala/dpla/ingestion3/utils/Utils.scala
mdellabitta: In general, they want you to use one Logger per class and use the Appender to talk about where things go. 

This is going to get complicated though when/if we use multiple machines to do this work, it's not going to magically go over the wire back to the master system. And I'm not sure stuff in System properties is distributed either. 

If we really wanted to avoid logging to Spark's logs, probably we would need the information to instantiate an Appender to be serialized over the wire, and the Loggers to be created from that. And then at job wrap up, we'd need something to go collect the logs.

In the long run, it might actually be easier to just piggyback on Spark's Appender and deal with the merged streams, because we'd be reinventing this wheel in a distributed context (YARN and EMR do log aggregation and archiving for you.)

But right now, we're not distributed, so let's just keep this.

moltude:
@mdellabitta I think I've addressed all the issues you raised in your review.

moltude:
Thanks all -- merging shortly

PR #79, opened 2017-09-14
Avro Storage Integration -- DO NOT MERGE

mdellabitta:
This is a working implementation of getting rid of Kyro serialization in favor of native Avro serialization. Not ready for merge in that it will affect work in flight, will push through merges at the end of the sprint once dust settles.

### DIFF
--- a/src/main/scala/dpla/ingestion3/MappingEntry.scala
+++ b/src/main/scala/dpla/ingestion3/MappingEntry.scala
@@ -2,23 +2,28 @@ package dpla.ingestion3
 
 import java.io.File
 
-import dpla.ingestion3.mappers.providers.{CdlExtractor, NaraExtractor, PaExtractor}
-import dpla.ingestion3.model.{DplaMap, DplaMapData, DplaMapError}
+import dpla.ingestion3.mappers.providers.{CdlExtractor, Extractor, NaraExtractor, PaExtractor}
+import dpla.ingestion3.model.{DplaMapData, RowConverter}
 import dpla.ingestion3.utils.Utils
 import org.apache.log4j.LogManager
 import org.apache.spark.SparkConf
-import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.{DataFrame, Dataset, Row, SparkSession}
 import java.io._
+
+import com.databricks.spark.avro._
+import org.apache.spark.sql.catalyst.encoders.{ExpressionEncoder, RowEncoder}
+
+import scala.util.{Failure, Success}
 
 /**
   * Expects two parameters:
-  *   1) a path to the harvested data
-  *   2) a path to output the mapped data
+  * 1) a path to the harvested data
+  * 2) a path to output the mapped data
   *
-  *   Usage
-  *   -----
-  *   To invoke via sbt:
-  *     sbt "run-main dpla.ingestion3.MappingEntry /input/path/to/harvested/ /output/path/to/mapped/"
+  * Usage
+  * -----
+  * To invoke via sbt:
+  * sbt "run-main dpla.ingestion3.MappingEntry /input/path/to/harvested/ /output/path/to/mapped/"
   */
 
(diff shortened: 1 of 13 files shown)


src/main/scala/dpla/ingestion3/MappingEntry.scala
```diff
@@ -71,36 +78,34 @@ object MappingEntry {
...
-    // TODO there is probably a much cleaner/better way of writing this.
-    val mappingSuccess = mappingResults
-      .filter(r => r.isInstanceOf[DplaMapData])
-      .map(r2 => r2.asInstanceOf[DplaMapData])
-
-    // TODO ditto ^^ This converts the Dataset to an Array[String], This is an easy way out for now.
-    val failures = mappingResults
-      .filter(r => r.isInstanceOf[DplaMapError])
-      .map(r2 => r2.asInstanceOf[DplaMapError]).map(f => f.errorMessage).collect()
+    val documents = harvestedRecords.select("document").as[String]
```
AudreyAltman: I think that in the line above, `documents` is a `Dataset[String]` whereas `harvestedRecords` is a `DataFrame`?  Is that correct?  Maybe would be nice to specify when declaring the `val`'s.

src/main/scala/dpla/ingestion3/MappingEntry.scala
```diff
@@ -71,36 +78,34 @@ object MappingEntry {
...
-    mappingSummary(harvestedRecords.count(), mappingSuccess.count(),failures, dataOut, shortName)
+    mappingSummary(harvestedRecords.count(), successResults.count(), failures, dataOut, shortName)
+
+    spark.stop()
   }
 
+  private def map(extractorClass: Class[_ <: Extractor], document: String): (Row, String) =
+    extractorClass.getConstructor(classOf[String]).newInstance(document).build() match {
+      case Success(dplaMapData) => (RowConverter.toRow(dplaMapData, model.sparkSchema), null)
+      case Failure(exception) => (null, exception.getMessage)
```
AudreyAltman: What is the reason for using `null` in the two lines above, as opposed to `Option`?

mdellabitta: I could be wrong, but I had trouble dealing with Option in Dataframes and Datasets. They just aren't meant for that. I don't know that I tried it in this particular code, so I'll try it quickly to make sure.

mdellabitta: This seems to back up what I have found so far, that it's non-trivial: https://stackoverflow.com/questions/36449368/how-to-create-encoder-for-option-type-constructor-e-g-optionint

src/main/scala/dpla/ingestion3/MappingEntry.scala
moltude: @AudreyAltman concur

src/main/scala/dpla/ingestion3/MappingEntry.scala
```diff
@@ -71,36 +78,34 @@ object MappingEntry {
...
+    val mappingResults = documents.map(document => map(extractorClass, document))(tupleRowStringEncoder)
 
     // Delete the output location if it exists
     Utils.deleteRecursively(new File(dataOut))
 
-    // Save successfully mapped records out to Avro file
-    mappingSuccess.toDF("document").write
-      .format("com.databricks.spark.avro")
-      .save(dataOut)
+
```
moltude: Well this cleaned up nicely ;)

src/main/scala/dpla/ingestion3/MappingEntry.scala
AudreyAltman: Ah, that makes sense.

mdellabitta: It does, only because they need to give us more options for encoding... Options.

PR #80, opened 2017-09-15
Add Minnesota mapping

moltude:
Add the MDLExtractor class and update the MappingEntry with a case for 'mdl'


moltude:
```Harvested 547051 records
Mapped 546057 records
Failed to map 994 records.```

Errors are records that are missing dataProvider value or have invalid characters in URIs

PR #82, opened 2017-09-15
WI mapping

moltude:
Adds WI extractor and mapping entry point.

This is a lingering branch from a prior sprint that didn't get a PR.

### DIFF
--- a/src/main/scala/dpla/ingestion3/MappingEntry.scala
+++ b/src/main/scala/dpla/ingestion3/MappingEntry.scala
@@ -1,6 +1,6 @@
 package dpla.ingestion3
 
-import java.io.File
+import java.io.{File, PrintWriter}
 
 import dpla.ingestion3.mappers.providers._
 import dpla.ingestion3.model.{DplaMapData, RowConverter}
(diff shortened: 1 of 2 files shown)


moltude:
wth -- why only one check run here?

src/main/scala/dpla/ingestion3/MappingEntry.scala
```diff
@@ -1,20 +1,19 @@
...
 import dpla.ingestion3.model.{DplaMapData, RowConverter}
 import dpla.ingestion3.utils.Utils
 import org.apache.log4j.LogManager
 import org.apache.spark.SparkConf
 import org.apache.spark.sql.{DataFrame, Dataset, Row, SparkSession}
-import java.io._
-
-import com.databricks.spark.avro._
 import org.apache.spark.sql.catalyst.encoders.{ExpressionEncoder, RowEncoder}
+import com.databricks.spark.avro._
```
moltude: Needed this to get line 64 to compile

src/main/scala/dpla/ingestion3/mappers/providers/WiExtractor.scala
```diff
@@ -0,0 +1,78 @@
...
+      ),
+
+      EdmWebResource(
+        // TODO is this the correct mapping for uri? What about OreAgg.`object`?
+        uri = itemUri
+      ),
+
+      OreAggregation(
+        uri = mintDplaItemUri(),
+        //below will throw if not enough contributors
```
AudreyAltman: Do you want it to throw?  If this is something we need to come back to later, maybe make this comment at "TODO" so we don't forget about about?

PR #83, opened 2017-09-21
Add Twofishes hostname configuration to application configuration

moltude:
Addresses https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1591

* Refactor of HarvesterCmdArgs to CmdArgs and made --input parameter optional to support harvests.
* Refactor of EnrichEntry and tests to use the configurations from Ingesiton3Conf
* Update sample configuration file

### DIFF
--- a/src/main/scala/dpla/ingestion3/confs/Ingestion3Conf.scala
+++ b/src/main/scala/dpla/ingestion3/confs/Ingestion3Conf.scala
@@ -3,7 +3,17 @@ package dpla.ingestion3.confs
 import com.typesafe.config.ConfigFactory
 import org.rogach.scallop.{ScallopConf, ScallopOption}
 
-class Ingestion3Conf(confFilePath: String, providerName: String) extends ConfUtils {
+/**
+  *
+  * @param confFilePath Required for all operations (harvest, mapping,
+  *                     or enrichment)
+  * @param providerName Optional - Provider shortName used to lookup provider
+  *                     specific settings in application configuration file.
+  *
+  *                     Harvest operations require a set of provider settings
+  *
+  */
+class Ingestion3Conf(confFilePath: String, providerName: Option[String] = None) extends ConfUtils {
   def load(): i3Conf = {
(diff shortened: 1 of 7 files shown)


src/main/scala/dpla/ingestion3/confs/Ingestion3Conf.scala
```diff
@@ -15,18 +25,16 @@ class Ingestion3Conf(confFilePath: String, providerName: String) extends ConfUti
...
 
-    val providerConf = ConfigFactory
-      .load
-      .getConfig(providerName)
-      .withFallback(baseConfig)
-      .resolve()
+    val providerConf = providerName match {
+      case Some(name) => ConfigFactory.load
+        .getConfig(name)
+        .withFallback(baseConfig)
```
AudreyAltman: With this fallback to `baseConfig` - is this just an empty config, or are there default values?  And if there are default values, where are they defined?

moltude: `baseConfig` includes all the non-provider specific settings (`spark`, `twofishess` etc.) defined within the application configuration. 

Properties set outside of a provider context can be overridden within a provider block so those are the "defaults" in some sense but the absolute fallback (currently) is accessing the configs via `getOrElse` in the Entry/driver classes. For example: 

`.setMaster(i3Conf.spark.sparkMaster.getOrElse("local[*]"))`

moltude: Because the `required` properties are different for each operation it didn't make sense to me to do any "generic" validation of the config file. Rather push that validation down into the Entry point classes. 

The twofishes host is only required for the enrichment jobs and even then only if you are running Spatial so I didn't want to throw errors if it wasn't included and we tried to run a harvest job.

AudreyAltman: :+1:

PR #84, opened 2017-09-21
Single entry point for all harvesters

AudreyAltman:
This creates a single entry point for harvesters.

Here is a breakdown of the new objects/classes:
- `HarvestEntry` - the main entry point for all harvesters.
-  `Harvester` - abstract class that takes care of logic common among all harvesters
    - `OaiHarvester` - implementation of `Harvester`
    - `PssHarvester` - implementation of `Harvester`
    - `ApiHarvester` - abstract class that takes care of logic common among all API harvesters
        - `CdlHarvester` - implementation of `ApiHarvester`
        - `MdlHarvester` - implementation of `ApiHarvester`

I used abstract classes in instead of traits so that I could set instance variables.  Things like the configs and loggers got used all over the place, and it proved quite useful to set them as instance vars rather than passing them like hot potatoes between methods.

I have tested this with IL, MDL, CDL, and PSS and verified that the data output is the same.  We decided to put off the resource sync harvester until it has been refactored.

One thing that may need followup - The OAI and PSS harvesters save out a directory containing multiple avro files, while the CDL and MDL harvesters save out a single avro file.  Likewise, the `--output` command line arg is the name of a directory for OAI and PSS (ie. "my-harvest"), while it must have an `.avro` file extension for MDL and CLD (ie. "my-harvest.avro").  Is this inconsistency going to to cause us trouble later on?


src/main/scala/dpla/ingestion3/harvesters/Harvester.scala
```diff
@@ -0,0 +1,145 @@
...
+    * Initiate a spark session using the configs specified in the i3Conf.
+    *
+    * Lazy evaluation b/c some harvesters do not need a spark context until the
+    * very end of the harvest.
+    *
+    * @return SparkSession
+    */
+  protected lazy val spark: SparkSession = {
+    val sparkConf = new SparkConf().setAppName(s"Harvest: ${provider}")
+    sparkConf.setMaster(conf.spark.sparkMaster.get)
```
moltude: Should this be getOrElse("local[1]") or such? The conf file isn't doing any validation of its contents or do we want to throw an exception if values aren't specified in the conf file?

src/main/scala/dpla/ingestion3/harvesters/api/CdlHarvester.scala
```diff
@@ -17,9 +21,66 @@ import scala.util.{Failure, Success, Try}
...
+            continueHarvest = false
+        }
+      case _ =>
+        harvestLogger.error("Harvest returned None")
+        continueHarvest = false
+    }
+  }
 
   /**
     * Get a single-page, un-parsed response from the OAI feed, or an error if
```
moltude: Noticing my own mistake here: `CDL` not `OAI` feed

AudreyAltman: Good catch!

src/main/scala/dpla/ingestion3/harvesters/Harvester.scala
AudreyAltman: Yeah, that's a good idea.  I'll set a default.

moltude:
@AudreyAltman That last point about `.avro` vs directory output has been bothering me for a while too. I think we should just pick one and stick to it. I'm in favor of using folders and losing the `.avro`s

mdellabitta:
Also I agree that we should be directory rather than file based. It just works in this Spark world a bit better.

moltude:
@mdellabitta re:logging -- If we have per class loggers does that mean we still send all output to the same log file but at different levels for a given class?

AudreyAltman:
@mdellabitta it wouldn't be hard for me to fix the loggers so that it's one per class.  I can do that before we merge.

mdellabitta:
Yeah, Appenders are the level at which you talk about logfiles, Loggers are the API for a class to write stuff to $somewhere. The config file or runtime programmatic configuration knits them together and says that Logger X logs all messages say INFO and above to Appender X and DEBUG and above to Appender Y.

Unfortunately, I know of nothing that makes this all work in distributed context, so if we go that route, the single log file idea probably has to go away.

@AudreyAltman probably don't worry about it at this point, because without a deeper dive, you'd break the Appender part that makes the unified log file that @moltude built.

AudreyAltman:
Ah I see.  Okay I will leave it then.  Thanks for the explanation @mdellabitta

AudreyAltman:
@moltude and @mdellabitta one more question - in the final DataFrame that we save out, should we be saving the full provider name (e.g. "California Digital Library") or the provider short name (e.g. "cdl")?

mdellabitta:
I think the short name.

moltude:
short name

AudreyAltman:
I made a ticket to address the directory structure of saved files so this PR doesn't get overloaded: https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1607

moltude:
@AudreyAltman 👍

PR #85, opened 2017-09-22
Refactor around making OreAggregation the top level entity

mdellabitta:
- Makes OreAggregation the top level entity in the data model. 
- DplaSourceResource is now a child of OreAggregation. 
- The previously top level EdmWebResource is now isShownAt.
- Adds dplaUri field to OreAggregation as the place to store the DPLA URI

### DIFF
--- a/src/main/scala/dpla/ingestion3/model/DplaMapData.scala
+++ b/src/main/scala/dpla/ingestion3/model/DplaMapData.scala
@@ -8,32 +8,34 @@ import dpla.ingestion3.model.DplaMapData._
   * Contains type definitions that express cardinality of fields
   * and a union type for fields that can be literals or uris.
   */
-
-
 object DplaMapData {
   type ZeroToMany[T] = Seq[T]
   type AtLeastOne[T] = Seq[T]
   type ZeroToOne[T] = Option[T]
   type ExactlyOne[T] = T
-  type LiteralOrUri = Either[String,URI]
+  type LiteralOrUri = Either[String, URI]
 }
 
-/**
-  * Container for the classes that represent an item in DPLA MAP.
-  *
-  * @see https://dp.la/info/developers/map/
-  *
-  * @param sourceResource Metadata about the source item rather than it's representations.
-  * @param edmWebResource Metadata about the item's representation on the provider's site.
-  * @param oreAggregation Metadata about the aggretation of the item across it's source and representations.
-  */
-case class DplaMapData (
-                         sourceResource: DplaSourceResource,
-                         edmWebResource: EdmWebResource,
-                         oreAggregation: OreAggregation
-                       )
-
 //Core Classes
+
+/**
+  * The aggregation of attributes that apply to the described resource as a whole,
+  * grouped from edm:WebResource and dpla:SourceResource.
+  *
+  */
+case class OreAggregation(
+                           dplaUri: ExactlyOne[URI],
+                           sourceResource: ExactlyOne[DplaSourceResource],
+                           dataProvider: ExactlyOne[EdmAgent],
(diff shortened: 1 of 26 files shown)


src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -134,30 +104,18 @@ case class SkosConcept(
...
-                            Esp. since in date in sourceResource is also ZeroToMany...
-
-                            I'm changing the type for the enrichment driver and need to follow-up
-                            w/team
-
-                            Also added a prefLabel property to store the enriched label
-                         */
                         originalSourceDate: ZeroToOne[String] = None,
-                        prefLabel: ZeroToOne[String] = None,
+                        prefLabel: ZeroToOne[String] = None, //todo not in spec?
```
moltude: I thought we wanted a pref label for our enrichment generated date labels while preserving their original label?  It may not be in the spec but IIRC that was the change we discussed. Whether we should get ahead of the spec here is another question but the intent was clear AFAIK.

src/main/scala/dpla/ingestion3/model/RowConverter.scala
```diff
@@ -16,103 +17,102 @@ import org.apache.spark.sql.types.{ArrayType, StringType, StructField, StructTyp
...
-  private[model] def edmAgent(ea: EdmAgent): Row = Row(
-    ea.uri.map(_.toString).orNull,
-    ea.name.orNull,
-    ea.providedLabel.orNull,
-    ea.note.orNull,
-    ea.scheme.map(_.toString).orNull,
-    ea.exactMatch.map(_.toString),
-    ea.closeMatch.map(_.toString)
+  private[model] def fromEdmAgent(ea: EdmAgent): Row = Row(
+    ea.uri.map(_.toString).orNull, //0
```
moltude: Thank you for commenting the array indices here. Future Scott owes you a beer for this stroke of genius.

mdellabitta:
Agreed... big PRs are hard to review.

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
mdellabitta: I actually don't remember the conversation, actually. Just left a note to ask about it.

mdellabitta: So I thought about it, and I guess what I was wondering was, is our "enriched" model of a date EDTF, and represented in total by the begin and end fields, and a string-only representation redundant. My question is answered by the prefLabel field in MAP 4.1, so nevermind. :)

mdellabitta:
Rebased on develop, removed unneeded comment.

src/main/scala/dpla/ingestion3/model/ModelConverter.scala
```diff
@@ -14,13 +14,20 @@ import org.apache.spark.sql.Row
   * is responsible for converting a DplaMapData to a Row and represents the opposite of this transformation.
```
AudreyAltman: Do you want to change the comment block above to say `OreAggregation` instead of `DplaMapData`?

src/main/scala/dpla/ingestion3/reports/MetadataCompletenessReport.scala
```diff
@@ -85,7 +85,7 @@ class MetadataCompletenessReport(
   /**
     * Map a Dataset of DplaMapData to a Dataset of CompletenessTally.
```
AudreyAltman: Again, change comment above from `DplaMapData` to `OreAggregation`?

PR #88, opened 2017-09-26
Initial support for using Amazon S3

moltude:
* Refactor Ingestion3Conf to support loading configuration
  file from local filesystem and S3. S3 support requires AWS
  keys to be set ENV props, JAVA props, .aws/credentials or EC2
  metadata (see DefaultAWSCredentialsProviderChain)
* Add support for saving harvester output to AWS S3 for all
  Harvesters by setting AWS S3 access and secret key values within
  SparkContext hadoopConfiguration

### DIFF
--- a/src/main/scala/dpla/ingestion3/harvesters/Harvester.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/Harvester.scala
@@ -2,12 +2,13 @@ package dpla.ingestion3.harvesters
 
 import java.io.File
 
+import com.amazonaws.auth.{AWSCredentialsProviderChain, DefaultAWSCredentialsProviderChain}
 import dpla.ingestion3.confs.i3Conf
 import dpla.ingestion3.utils.Utils
 import org.apache.log4j.Logger
-import org.apache.spark.{SparkConf, SparkContext}
 import org.apache.spark.sql.types._
 import org.apache.spark.sql.{DataFrame, SparkSession}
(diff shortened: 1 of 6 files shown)


src/main/scala/dpla/ingestion3/harvesters/Harvester.scala
```diff
@@ -52,6 +53,7 @@ abstract class Harvester(shortName: String,
   def harvest = {
 
     // If the output directory already exists then delete it and its contents.
+    // TODO Does this need to work with s3?
```
mdellabitta: I think we should stop automagically deleting our data. Particularly from S3. If you want to do it locally to be able to iterate quickly while developing, it's probably better done with a shell script that does it and then invokes the code you want to run again.

src/main/scala/dpla/ingestion3/harvesters/oai/OaiHarvester.scala
```diff
@@ -51,13 +50,16 @@ class OaiHarvester(shortName: String,
       .withColumn("provider", lit(shortName))
       .withColumn("mimetype", lit(mimeType))
 
+    harvestLogger.info(s"Saving output to ${outputDir}")
+
     // Write harvested data to file.
-    finalData.write
+    finalData
+      .write
+      .mode(SaveMode.Overwrite) // TODO do we want to overwrite, fail or make configurable?
```
mdellabitta: I think we want to fail.

PR #89, opened 2017-09-26
MappingEntry configuration

moltude:
Addresses DT-1612. Update MappingEntry to use standard command line args of:
--input
--output
--conf
--name

Reads Spark conf options from application.conf


src/main/scala/dpla/ingestion3/MappingEntry.scala
```diff
@@ -57,10 +87,7 @@ object MappingEntry {
 
     // Load the harvested record dataframe
     val harvestedRecords: DataFrame = spark.read.avro(dataIn)
-
-    // Take the first harvested record in the dataFrame and get the provider value
```
moltude: This is removed in favor of passing in the shortName via command line args so the provider-specific properties for Spark can be loaded before creating the SparkSession

moltude:
This was pretty straightforward so I'm going to go ahead and merge it.

PR #90, opened 2017-09-26
Fix/id minting inversion

moltude:
Addresses DT-1594. 

Moves the construction of the base id (what DPLA hashes) into the Extractor trait and adds getter methods to be implemented in the provider's extractor.
 
`useProviderName(): Boolean`
`getProviderId(): String`
`getProviderName(): String`

These methods provide the Extractor trait with all it needs to create the base id, mint the DPLA id and construct the DPLA item URI.  The provider extractor now explicitly states what the base ID is to be constructed from and Extractor does the construction.

If the provider's identifier cannot be or if the combination of values for (useProviderName and getProviderName().isEmpty) do not make sense (e.g. false, false or true, true) extracted an exception is raised and the record is not mapped.

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
@@ -2,8 +2,8 @@ package dpla.ingestion3.mappers.providers
 
 import java.net.URI
 
-import dpla.ingestion3.model.{DplaMapData, EdmAgent, OreAggregation}
-import dpla.ingestion3.utils.Utils
+import dpla.ingestion3.model.{EdmAgent, OreAggregation}
+import org.apache.commons.codec.digest.DigestUtils
 
 import scala.util.{Failure, Success, Try}
 
@@ -13,29 +13,88 @@ import scala.util.{Failure, Success, Try}
 
 trait Extractor {
   // Base item uri
-  private val baseItemUri = "http://dp.la/api/items/"
+  private val baseDplaItemUri = "http://dp.la/api/items/"
 
   def build(): Try[OreAggregation]
   def agent: EdmAgent
+
   /**
-    * Build the base ID to be hashed. Implemented per provider
+    * Does the provider use a prefix (typically their provider abbreviation) to
+    * salt the identifier?
+    *
+    * @return Boolean
+    */
+  def useProviderName(): Boolean
+
+  /**
+    * Extract the record's "persistent" identifier. Implementations should raise
+    * an Exception if no ID can be extracted
+    *
+    * @return String Record identifier
+    * @throws Exception If ID can not be extracted
+    */
+  @throws(classOf[ExtractorException])
+  def getProviderId(): String
+
+  /**
+    * The value to salt the identifier with. Not all providers
+    * use this. If not, an empty string should be used
+    * @return
+    */
+  def getProviderName(): String
+
+  /**
+    * Builds the ID to be hashed by either concatenating the provider's
+    * abbreviated name and the persistent identifier for the record with
+    * a double dash `--` or only the persistent identifier.
+    *
+    * Some providers use a name prefix and some do not. Please see Hub
+    * mapping documentation and individual Extractor classes.
+    *
+    * ############################################################
+    * # WARNING DO NOT CHANGE UNLESS YOU KNOW WHAT YOU ARE DOING #
+    * ############################################################
     *
     * @return String
     */
-  def getProviderBaseId(): Option[String]
+  def buildProviderBaseId(): String = {
+    def idErrorMsg(): String = {
+      s"Unable to mint ID given values of:\n" +
+        s"useProviderName: ${useProviderName()}\n" +
+        s"getProviderName: ${getProviderName()}\n" +
+        s"getProviderId: ${getProviderId()}\n"
+    }
+    Try {
+      useProviderName() match {
+        // use prefix of provider short name
+        case true => s"${getProviderName()}--${getProviderId()}"
+        // do not use prefix
+        case false => getProviderId()
+      }
+    } match {
+      case Success(id) => id
+      case Failure(_) => throw ExtractorException(idErrorMsg())
+    }
+
+  }
 
   /**
     * Hashes the base ID
     *
-    * @return MD5 hash of the base ID
+    * ############################################################
+    * # WARNING DO NOT CHANGE UNLESS YOU KNOW WHAT YOU ARE DOING #
+    * ############################################################
+    *
+    * @return String MD5 hash of the base ID
     */
-  protected def mintDplaId(): String = Utils.generateMd5(getProviderBaseId())
+  protected def mintDplaId(): String = DigestUtils.md5Hex(buildProviderBaseId())
 
   /**
     * Builds the item URI
     *
     * @return URI
     */
-  protected def mintDplaItemUri(): URI = new URI(s"${baseItemUri}${mintDplaId()}")
-
+  protected def mintDplaItemUri(): URI = new URI(s"$baseDplaItemUri${mintDplaId()}")
 }
+
+case class ExtractorException(message: String) extends Exception(message)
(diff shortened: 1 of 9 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
```diff
@@ -2,40 +2,87 @@ package dpla.ingestion3.mappers.providers
...
+        s"getProviderName: ${getProviderName()}"
+    }
+
+    (useProviderName(), getProviderName().isEmpty) match {
+      // use prefix and prefix is set
+      case (true, false) => s"${getProviderName()}--${getProviderId()}"
+      // do not use prefix and prefix is not set
+      case (false, true) => getProviderId()
+      // use prefix but prefix is not set
+      case (true, true) => throw new Exception(idErrorMsg())
```
moltude: These last two cases could be collapsed into `case _` but I thought it would better to be as explicit as possible here. Would specific error messages for each case be useful?

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
```diff
@@ -2,40 +2,88 @@ package dpla.ingestion3.mappers.providers
...
+    * Does the provider use a prefix (typically their provider abbreviation) to
+    * salt the identifier?
+    *
+    * @return Boolean
+    */
+  def useProviderName(): Boolean
+
+  /**
+    * Extract the record's "persistent" identifier. Implementations should raise
+    * an Exception if no ID can be extracted
```
AudreyAltman: Can you add an `@throws` comment here to reflect the expectation an an exception may be thrown?  I don't know if that's a good idea or not, just a thought.

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
```diff
@@ -2,40 +2,88 @@ package dpla.ingestion3.mappers.providers
...
+    *
+    * @return String Record identifier
+    */
+  def getProviderId(): String
+
   /**
-    * Build the base ID to be hashed. Implemented per provider
+    * The value to salt the identifier with. Not all providers
+    * use this. If not, an empty string should be used
```
AudreyAltman: What's the benefit to using an empty string instead of following a Some/None model?

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
AudreyAltman: I think it's good the way you wrote it.

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
moltude: we got into trouble passing in options because they weren't getting decomposed and we ended up hashing Some(”id") rather than "id".  When choosing this method I was trying to be as explicit as possible all the way through

I'll take another look at it though to see if there might be another way...

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
moltude: we got into trouble passing in options because they weren't getting decomposed and we ended up hashing Some(”id") rather than "id".  When choosing this method I was trying to be as explicit as possible all the way through

I'll take another look at it though to see if there might be another way...

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
moltude: Done.

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
mdellabitta: I was thinking that every provider definitely should have a shortname, but `useProviderName()` would determine if it were used in the ID minting.

mdellabitta: The reason being that the shortname could be useful in other contexts, like automatically building a registry of Extractor implementations, for example.

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
```diff
@@ -2,40 +2,91 @@ package dpla.ingestion3.mappers.providers
...
-  protected def mintDplaId(): String = Utils.generateMd5(getProviderBaseId())
+  protected def mintDplaId(): String = DigestUtils.md5Hex(buildProviderBaseId())
 
   /**
     * Builds the item URI
     *
     * @return URI
     */
-  protected def mintDplaItemUri(): URI = new URI(s"${baseItemUri}${mintDplaId()}")
+  protected def mintDplaItemUri(): URI = new URI(s"$baseDplaItemUri${mintDplaId()}")
```
mdellabitta: I'm thinking a DO NOT TOUCH UNLESS YOU REALLY KNOW WHAT YOU'RE DOING- style warning message might be appropriate for this method and the one above it

moltude: Good call.

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
moltude: @mdellabitta If that is the case should it read it out of the config file or is that a vector for unreviewed changes that would impact on IDs that we want to actively prevent?

moltude: Those provider `shortName` are first set by command line args in the HarvesterEntry and then embedded into the harveted DataFrame.  So if we want to rely on those we **could** have an issue where we change the shortName during invocation b/c they changed names w/o realizing that it is going to disrupt the ID minting.  I don't disagree with making every provider have a short name value set for ID minting and relying on the bool for construction.

I'm trying to argue for: 
1) Clustering all the aspects of an ID together 
2) Hard-coding those values as much as possible so any change would trigger a review.

mdellabitta: I guess I'm trying to get us to have one place to look when we decide what the shortname of a provider should be. Not having this be standardized is biting us in a number of places in the previous ingestion systems.

The problem you're bringing up is actually another iteration of this general problem. I think we should be able to circumvent  by having somewhere in a constant a list of all the Extractors that are valid, and therefore a list of the valid shortnames is easily obtainable. Then if anything was invoked with an invalid shortname, we should be able to instantly throw up on it, in any entry point.

moltude: Okay I see what you are saying. We can look to this block to do that validation of the shortName before we even hit the Extractor. There is probably a better way to achieve what you are describing but functionally I think it does exactly what we would want ATM. 

https://github.com/dpla/ingestion3/blob/661c0e33154dee45a343bcac6031e6a1b8802e36/src/main/scala/dpla/ingestion3/MappingEntry.scala#L64-L74

mdellabitta: Right, I'm going for the DRY approach.

You can see how we've already messed this up with PA, which I seem to remember you fixed in one of your many PRs today. Also we should be able to do this validation in the HarvestEntry as well.

moltude: it is in this PR ;)

moltude:
The last commit addresses issues raised by Michael about requiring a provider shortName for all providers and then relying only on `useProviderName` when minting an id.

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
```diff
@@ -60,23 +64,27 @@ trait Extractor {
...
-      // use prefix and prefix is set
-      case (true, false) => s"${getProviderName()}--${getProviderId()}"
-      // do not use prefix and prefix is not set
-      case (false, true) => getProviderId()
-      // use prefix but prefix is not set
-      case (true, true) => throw new Exception(idErrorMsg())
-      // do not use prefix but prefix value is set
-      case (false, false) => throw new Exception(idErrorMsg())
-      case _ => throw new Exception(idErrorMsg())
+    Try {
```
moltude: This Try{} block is probably excessive.

src/main/scala/dpla/ingestion3/mappers/providers/Extractor.scala
mdellabitta: Ha!

PR #91, opened 2017-09-27
save all harvests in consistent directory structure

AudreyAltman:
This changes the file structure of saved API harvests to be consistent with other harvests.  With this implementation, the `--output` command line param for all harvesters should be the name of a directory that will hold `.avro` files.  The API harvester creates the directory and the file within.

This has been tested locally with an MDL harvest.  It addresses [DT-1607](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1607).

### DIFF
--- a/src/main/scala/dpla/ingestion3/harvesters/api/ApiHarvester.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/api/ApiHarvester.scala
@@ -41,7 +41,7 @@ abstract class ApiHarvester(shortName: String,
   protected lazy val avroWriter: DataFileWriter[GenericRecord] = {
     val dirName = if (outputDir.endsWith("/")) outputDir.dropRight(1) else outputDir
     val fileName = dirName + "/api_harvest.avro"
-    val dir = new File(dirName).mkdir
+    val dir = new File(dirName).mkdirs
     val file = new File(fileName)
     getAvroWriter(file, schema)
   }
--- a/src/main/scala/dpla/ingestion3/harvesters/api/ApiHarvester.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/api/ApiHarvester.scala
@@ -1,5 +1,7 @@
 package dpla.ingestion3.harvesters.api
 
+import java.io.File
+
 import org.apache.avro.Schema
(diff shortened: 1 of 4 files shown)


moltude:
@AudreyAltman @mdellabitta Can `getAvroWriter` be moved out of `NaraHarvesterMain`?  It seems a bit awkward having a common method like that in a driver.

src/main/scala/dpla/ingestion3/harvesters/api/ApiHarvester.scala
```diff
@@ -36,8 +38,13 @@ abstract class ApiHarvester(shortName: String,
     * This is lazy b/c queryParams should be printed before avroWriter is set.
     * @see doHarvest
     */
-  protected lazy val avroWriter: DataFileWriter[GenericRecord] =
-  getAvroWriter(outputFile, schema)
+  protected lazy val avroWriter: DataFileWriter[GenericRecord] = {
+    val dirName = if (outputDir.endsWith("/")) outputDir.dropRight(1) else outputDir
+    val fileName = dirName + "/api_harvest.avro"
+    val dir = new File(dirName).mkdir
```
moltude: What about using `mkdirs` instead of `mkdir` to handle the case where multiple parent directories do not already exist?

AudreyAltman: Excellent point.

AudreyAltman:
@moltude I think that it makes sense to move `getAvroWriter` out of `NaraFileHarvesterMain`, if @mdellabitta agrees.  Do you want to make a ticket for that?

mdellabitta:
I agree that it should be done. For five minute things like this (if people agree it's a five minute thing) we should probably just do it rather than make a ticket.

http://media.riffsy.com/images/7cd63e804a14789acef74bc1918f7a08/tenor.gif

AudreyAltman:
Awesome, I nominate @moltude to do it b/c I have no intuition about where it should go.

moltude:
On it

moltude:
fixed up and rebased. @AudreyAltman merge when ready?

AudreyAltman:
Thanks @moltude

PR #99, opened 2017-10-03
Fixup match

moltude:
No-op if twofishes is reachable.


mdellabitta:
I usually just stick `Unit` in there when I mean "no-op"

PR #100, opened 2017-10-05
Improve EDTF handling

markbreedlove:
See https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1616

The changes here add what I think is EDTF Level 0 support.  See the tests. See https://www.loc.gov/standards/datetime/pre-submission.html (and note that the tests check for the example patterns in that document).

Though I've added the code to make sure Level 0 is handled, I haven't _removed_ old code that may be redundant. DT-1616 is a timebox, and I'm at my time limit.

For discussion:
1. Code quality. E.g. the embedded matches in `timeSpanFromEDTF` ... there is probably a better way. I could not get a chain of `getOrElse` to work.
2. What needs to be removed now?

I made the new methods I added public because I was having trouble testing them with `PrivateMethodTester`.

### DIFF
--- a/src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
+++ b/src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
@@ -15,6 +15,12 @@ class ParseDateEnrichment extends Serializable {
     val dateString = edtfOriginalDate.originalSourceDate.getOrElse("")
     val str = preprocess(dateString)
 
+
+    timeSpanFromEDTF(str) match {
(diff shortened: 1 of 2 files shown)


mdellabitta:
So the intent was that this was going to be in a standalone class, and would be called in the line where it says `//todo full EDTF parsing`, not rolled into the ParseDateEnrichment inline. Is it possible to extract this to it's own class? 

The idea is that we should try parsing the date as a standard EDTF date, and fall back to other heuristics if that doesn't work.

markbreedlove:
As  suggested, I'll merge this after squashing.

PR #101, opened 2017-10-06
Use accumulators for stats & add stack traces to error messages

markbreedlove:
See https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1617

Use Spark accumulators to track counts of total records processed and records successfully processed, instead of using `count()` methods.

Add stack trace strings to exception messages in `MappingEntry` and `EnrichEntry`.

### DIFF
--- a/src/main/scala/dpla/ingestion3/EnrichEntry.scala
+++ b/src/main/scala/dpla/ingestion3/EnrichEntry.scala
@@ -71,6 +71,7 @@ object EnrichEntry {
     val sc = spark.sparkContext
     val totalCount: LongAccumulator = sc.longAccumulator("Total Record Count")
     val successCount: LongAccumulator = sc.longAccumulator("Successful Record Count")
+    val failureCount: LongAccumulator = sc.longAccumulator("Failed Record Count")
 
     // Need to keep this here despite what IntelliJ and Codacy say
     import spark.implicits._
@@ -87,7 +88,7 @@ object EnrichEntry {
       val driver = new EnrichmentDriver(i3Conf)
       Try{ ModelConverter.toModel(row) } match {
         case Success(dplaMapData) =>
-          enrich(dplaMapData, driver, totalCount, successCount)
+          enrich(dplaMapData, driver, totalCount, successCount, failureCount)
         case Failure(err) =>
           (null, s"Error parsing mapped data: ${err.getMessage}\n" +
                  s"${err.getStackTrace.mkString("\n")}")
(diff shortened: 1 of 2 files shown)


src/main/scala/dpla/ingestion3/EnrichEntry.scala
```diff
@@ -103,23 +110,31 @@ object EnrichEntry {
...
     failures.foreach(msg => enrichLogger.warn(s"Error: ${msg}"))
 
-    enrichLogger.debug(s"Mapped ${mappedRecordCount} records and enriched ${enrichedRecordCount} records")
-    enrichLogger.debug(s"${mappedRecordCount-enrichedRecordCount} enrichment errors")
+    enrichLogger.debug(
+      s"Mapped ${totalCount.value} records and enriched ${successCount.value}" +
+        s" records"
+    )
+    enrichLogger.debug(
+      s"${totalCount.value - successCount.value} enrichment errors"
```
mdellabitta: Probably makes sense to count failures outright so they show up in the Spark UI.

src/main/scala/dpla/ingestion3/EnrichEntry.scala
```diff
@@ -103,23 +110,31 @@ object EnrichEntry {
...
+                     successCount: LongAccumulator): (Row, String) = {
+    totalCount.add(1)
     driver.enrich(dplaMapData) match {
-      case Success(enriched) => (RowConverter.toRow(enriched, model.sparkSchema), null)
-      case Failure(exception) => (null, exception.getMessage)
+      case Success(enriched) =>
+        successCount.add(1)
+        (RowConverter.toRow(enriched, model.sparkSchema), null)
+      case Failure(exception) =>
+        (null, exception.getMessage)
```
mdellabitta: There's a bit in the story about changing this so the whole exception stack trace gets turned into a string rather than just the message.

mdellabitta: Ah looks like you did it in MappingEntry. Do it here as well?

src/main/scala/dpla/ingestion3/MappingEntry.scala
```diff
@@ -76,6 +77,10 @@ object MappingEntry {
       .config(sparkConf)
       .getOrCreate()
 
+    val sc = spark.sparkContext
+    val totalCount: LongAccumulator = sc.longAccumulator("Total Record Count")
+    val successCount: LongAccumulator = sc.longAccumulator("Successful Record Count")
```
mdellabitta: FailureCount here as well?

mdellabitta: NM, github's being weird.

PR #102, opened 2017-10-06
Add JValue field, sidecar, to OreAggregation

moltude:
Allows mapped records to carry additional information within a JValue object. We can be more nuaunced about
what we track without forcing everything into a (String,String) map. The JValue is converted to a String
in the RowConverter.

### DIFF
--- a/src/main/scala/dpla/ingestion3/model/ModelConverter.scala
+++ b/src/main/scala/dpla/ingestion3/model/ModelConverter.scala
@@ -4,6 +4,10 @@ import java.net.URI
 
 import dpla.ingestion3.model.DplaMapData.LiteralOrUri
 import org.apache.spark.sql.Row
+import org.json4s.JsonAST.{JNothing, JValue}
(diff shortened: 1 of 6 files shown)


src/main/scala/dpla/ingestion3/model/ModelConverter.scala
```diff
@@ -141,4 +144,7 @@ object ModelConverter {
   private[model] def uriSeq(row: Row, fieldPosition: Integer): Seq[URI] =
     stringSeq(row, fieldPosition).map(new URI(_))
 
+  private[model] def toOptionJvalue(row: Row, fieldPosition: Integer): Option[JValue] = {
```
mdellabitta: This shouldn't block merging, but JValues can be JNothing, which might be easier to deal with than an Option, since you can run all the same path walking `\` things on JNothing. They just don't return results.

moltude:
@mdellabitta suggestion implemented

mdellabitta:
Sweet.

PR #103, opened 2017-10-06
Add ProviderRegistry

AudreyAltman:
This is meant to address [DT-1622](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1622) but I don't know if it does what the ticket intends.  @moltude can you take a look and give me feedback?  I may have wandered off into la-la land on this one.

### DIFF
--- a/src/main/scala/dpla/ingestion3/utils/ProviderRegistry.scala
+++ b/src/main/scala/dpla/ingestion3/utils/ProviderRegistry.scala
@@ -1,0 +1,74 @@
+package dpla.ingestion3.utils
+
+import dpla.ingestion3.harvesters.Harvester
+import dpla.ingestion3.harvesters.api.{CdlHarvester, MdlHarvester}
+import dpla.ingestion3.mappers.providers._
+
+import scala.util.Try
+
+/**
+  * Main entry point for accessing a provider's associated classes based on its
+  * short name.
+  *
+  * @example ProviderRegistry.lookupExtractor("nara") => Success[ classOf[NaraExtractor] ]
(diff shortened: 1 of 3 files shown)


moltude:
@AudreyAltman Gut check here. Did you mean DT-1618?   DT-1622 was my sidecar metadata ticket.

src/main/scala/dpla/ingestion3/utils/ProviderRegistry.scala
```diff
@@ -0,0 +1,74 @@
...
+    */
+  def lookupHarvesterClass(short: String): Try[Class[_ <: Harvester]] = Try {
+    registry.getOrElse(short, noHarvesterException(short))
+      .harvesterClass.getOrElse(noHarvesterException(short))
+  }
+
+  private case class Register(extractorClass: Option[Class[_ <: Extractor]] = None,
+                              harvesterClass: Option[Class[_ <: Harvester]] = None)
+
+  private val registry: Map[String, Register] = Map(
```
moltude: Really nice.
http://gph.is/1skTTcz

moltude:
Realizing this is out of scope for this ticket but I do have some concerns about how errors are now getting logged in MappingEntry.map(). 

PA only has ~187 mapping errors but the log file is now over 4,000 lines long if we are including stack traces. While this format my be helpful to us we should be thinking about a more concise and human readable summary that we can pass to either Gretchen or the provider. 

Also, the accumulator counts are double what they should be. PA reported 
`Harvested 464464 records`
`Mapped 464090 records`
`Failed to map 374 records.`

moltude:
https://i.imgur.com/xrDZuNK.gif

PR #104, opened 2017-10-11
Fixup Row > Model for sidecar

moltude:
Parsing an empty string was throwing errors when trying to convert Row > Model


moltude:
Thanks @AudreyAltman!

PR #105, opened 2017-10-12
Fix double-counting by Spark accumulators

markbreedlove:
See https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1640

This changeset uses Spark checkpoints and `persist()` calls to keep the dataframe from being scanned multiple times.

Can someone confirm the sanity of the creation of the checkpoint directory with a local file path, `/tmp/checkpoint`? Is this created on the Spark driver node? Do we have to worry about it being in S3 in an EMR context?  Do the executors need to have access to it?


mdellabitta:
I think in an EMR context, it'd likely be in HDFS. I believe HDFS comes with a tmp directory out of the box, but I'm not 100% sure, and I'm not sure that the default user has write privileges. Is there no checkpoint directory set by default? I imagine we might want to make the execution environment be in charge of where it is.

markbreedlove:
There is no checkpoint directory set by default; I don't know about a tmp directory, specifically. I got an error message saying I had to set a checkpoint directory.

mdellabitta:
Can you put a TODO comment in there to put that in a config?

markbreedlove:
OK, @mdellabitta, just did that.

PR #106, opened 2017-10-12
Revisions to JSON-L model (PA QA)

moltude:
* Add `prehashId` and `dplaId` values to `sidecar` in PAExtractor 
* Uses `prehashId` and `dplaId` to populate `@id` and `_id` in the JSON-L document 
* Tries to prettify OriginalRecord for output but this is not 100%. Issues with CQA frontend
*  nameOnlyConcept() now populates the `providedLabel` property not `concept`
* Fixup subject and language mapping in jsonlRecord()
* Add documentation

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/PaExtractor.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/PaExtractor.scala
@@ -5,13 +5,15 @@ import java.net.URI
 import dpla.ingestion3.mappers.xml.XmlExtractionUtils
 import dpla.ingestion3.model.DplaMapData.{ExactlyOne, ZeroToOne}
 import dpla.ingestion3.model._
+import dpla.ingestion3.utils.Utils
+import org.json4s.JsonDSL._
 
 import scala.util.Try
 import scala.xml._
 
 class PaExtractor(rawData: String, shortName: String) extends Extractor with XmlExtractionUtils {
 
-  implicit val xml: NodeSeq = XML.loadString(rawData)
+  implicit val xml: Elem = XML.loadString(rawData)
 
   // ID minting functions
   override def useProviderName(): Boolean = false
@@ -41,6 +43,8 @@ class PaExtractor(rawData: String, shortName: String) extends Extractor with Xml
   def build: Try[OreAggregation] = Try {
     OreAggregation(
       dplaUri = mintDplaItemUri(),
+      sidecar = ("prehashId", buildProviderBaseId()) ~
+                ("dplaId", mintDplaId()),
       sourceResource = DplaSourceResource(
         // This method of using NodeSeq is required because of namespace issues.
         collection = extractStrings(xml \ "metadata" \\ "relation").headOption.map(nameOnlyCollection).toSeq,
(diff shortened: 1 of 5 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/PaExtractor.scala
```diff
@@ -5,13 +5,15 @@ import java.net.URI
...
+import dpla.ingestion3.utils.Utils
+import org.json4s.JsonDSL._
 
 import scala.util.Try
 import scala.xml._
 
 class PaExtractor(rawData: String, shortName: String) extends Extractor with XmlExtractionUtils {
 
-  implicit val xml: NodeSeq = XML.loadString(rawData)
+  implicit val xml = XML.loadString(rawData)
```
mdellabitta: I get yelled at by IDEA when there's no type tag here because the field's not `private`. I assume you did this because you need a `Node` instead of a `NodeSeq` in order to format it in line 69?

src/main/scala/dpla/ingestion3/model/package.scala
```diff
@@ -105,7 +113,9 @@ package object model {
           ("identifier" -> record.sourceResource.identifier) ~
           ("language" ->
             record.sourceResource.language.map { lang =>
-              ("name" -> lang.providedLabel) ~ ("iso639_3" -> lang.concept)
+              "name" -> lang.concept.getOrElse(
+                          lang.providedLabel.getOrElse(""))
+               // FIXME what other SkosConcept fields do we need in the index for language?
```
mdellabitta: I think this is it, actually. :\

src/main/scala/dpla/ingestion3/model/package.scala
```diff
@@ -128,7 +138,9 @@ package object model {
            */
           // ("specType" -> "FIXME") ~
           // stateLocatedIn is omitted.
-          ("subject" -> record.sourceResource.subject.map{s => s.providedLabel}) ~
+          ("subject" -> record.sourceResource.subject.map{s =>
```
mdellabitta: Surprised anything hit the index before this without this fix.

src/main/scala/dpla/ingestion3/mappers/providers/PaExtractor.scala
moltude: Exactly but the formatting still isn't working so I can just change this back. Part of the problem is with the CQA app

src/main/scala/dpla/ingestion3/model/package.scala
moltude: 🎃

src/main/scala/dpla/ingestion3/model/package.scala
moltude: They'll hit the index but it won't render in the CQA app

src/main/scala/dpla/ingestion3/mappers/providers/PaExtractor.scala
moltude: I could also change it to `implicit val xml: Elem` I think that will get us both what we want atm.

moltude: @mdellabitta going with last option and then going to merge

mdellabitta: 👍

PR #108, opened 2017-10-12
Add EDTF Level 1 support to date enrichment

markbreedlove:
See https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1638
See https://www.loc.gov/standards/datetime/pre-submission.html

This adds Level 1 support and breaks out the EDTF processing to a new object, `dpla.ingestion3.utils.EDTFDate`.

See the tests in `EDTFDateTest`, which document exactly how everything is handled. Confirm that I've got everything in Level 1.

A few things are not being done yet:
* We're just truncating season data, because it is not well implemented in Level 1, which lacks the hemisphere syntax of Level 2. Seasons also have cultural relevance issues, as mentioned in the spec above.
* We're not doing exhaustive validation of dates that might be out of range for the given month, like `2017-11-31`.

### DIFF
--- a/src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
+++ b/src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
@@ -5,19 +5,27 @@ import java.text.SimpleDateFormat
 import scala.util.{Failure, Success, Try}
 import ParseDateEnrichment._
 import dpla.ingestion3.model.EdmTimeSpan
-
+import dpla.ingestion3.utils.EDTFDate
 import scala.annotation.tailrec
+
 
 class ParseDateEnrichment extends Serializable {
 
-  //TODO ranges
-  def parse(edtfOriginalDate: EdmTimeSpan, allowInterval: Boolean = false): EdmTimeSpan = {
-    val dateString = edtfOriginalDate.originalSourceDate.getOrElse("")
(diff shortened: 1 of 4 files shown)


src/main/scala/dpla/ingestion3/enrichments/ParseDateEnrichment.scala
```diff
@@ -5,19 +5,27 @@ import java.text.SimpleDateFormat
...
+    val str = preprocess(dateString)
 
-    timeSpanFromEDTF(str) match {
-      case Some(timespan) => return timespan
+    EDTFDate.rangeForEDTF(str) match {
+      case Some(range) =>
+        val begin = range._1 match { case "" => None case _ => Some(range._1)}
+        val end = range._2 match { case "" => None case _ => Some(range._2)}
+        return EdmTimeSpan(
+          originalSourceDate = Some(str),
```
mdellabitta: Should this be `dateString`?

markbreedlove: Oh, right, yes. Will change ...

markbreedlove:
Oh wow, I forgot something again. I'll make another commit. Though I added support for "open" intervals I didn't actually call `rangeForOpenInterval` from the `rangeForEDTF` method.

Sorry this is kind of a pattern, my issuing a PR then making changes.

PR #109, opened 2017-10-13
Add test coverage & send to Codacy

markbreedlove:
See https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1625


mdellabitta:
screen shot 2017-10-13 at 12 28 24 pm

Nice! Now all we have to do is make that number bigger. ;)

PR #110, opened 2017-10-13
Refactor and expand pre-mapping reports.

AudreyAltman:
This refactors pre-mapping reports.  There is now a trait called `Shredder` that holds all logic in common for JSON and XML shredders.  The class `PreMappingReport` runs several different calculations over shredded data.  `PreMappingReporter` handles user input and writes out reports.

In this implementation, running the pre mapping reporter process generates five different reports.  I'm not sure if we will need all these reports, and in the future we might want a more nuanced workflow where we can choose to generate only some reports.

This addresses [DT-1507](https://digitalpubliclibraryofamerica.atlassian.net/browse/DT-1507).  It has been tested locally with a sample of MDL data.


AudreyAltman:
The codacy error relates to the use of `null`, which I believe is unavoidable in this case.

src/main/scala/dpla/ingestion3/preMappingReports/PreMappingReporter.scala
```diff
@@ -0,0 +1,100 @@
...
+    writeElasticSearchData
+    writeCardinalityReport
+    writeOpenRefineData
+    writeRecordsMissingAttributesReport
+
+    // Stop spark context.
+    spark.sparkContext.stop()
+  }
+
+  private def writeShreddedData: Unit =
```
markbreedlove: IntelliJ IDEA is telling me that these methods of type `Unit` are supposed to have empty parentheses; e.g. `private def writeShreddedData(): Unit`, the convention being that there are parentheses if the method has side effects.

src/main/scala/dpla/ingestion3/preMappingReports/Shredder.scala
```diff
@@ -0,0 +1,35 @@
+package dpla.ingestion3.premappingreports
```
markbreedlove: IntelliJ IDEA is telling me that this package name does not match its directory name. (It must be because it's case-sensitive.) It says it could result in problems resolving dependencies.

src/main/scala/dpla/ingestion3/PreMappingReporterMain.scala
```diff
@@ -72,20 +42,25 @@ object PreMappingReporterMain {
...
-    val outputURI = args(1)
+    val input = args(0)
+    val output = args(1)
     val sparkMasterName = args(2)
     val token = args(3)
 
-    new PreMappingReporter(
-      token, inputURI, outputURI, sparkMasterName
-    ).main()
+    val result = new PreMappingReporter(input, output, sparkMasterName, token)
```
markbreedlove: Here in this file the last parameter and its variable are named `token` but elsewhere in `PreMappingReporter` it's referred to as `inputDataType`, which might be more clear, here and above as "input data type" in the usage documentation.

AudreyAltman:
Thanks @markbreedlove I've incorporated your feedback :)

PR #114, opened 2017-10-24
Initial single entry work

moltude:
Addresses DT-1672. This was a bit sprawling because of the work required to normalize the
executions of harvest, map, enrich with json-l and reporting. Json-l and reporting did not
have centralized logging or reading options from a config file. This should be tightend up.

* Move entries (harvest, map, enrich, json-l etc.) into "entries" package. Create seperate
  packages for ingest entries and report entries.
* Add IngestRemap entry point. Expects harvested data and runs mapping, enrichment
  json-l exports and metadata reports.
* Lots of minor refactoring around logging statements and messaging
* Tweak formatting of log messages in log4j.properties
* Refactor and try to standardize executor entries for harvest, mapping, enrichment, json-l
  and reports.
  - Add accessor methods to CmdArgs
  - Add createLogger() method in Utils
* Add executeAllReports() to ReporterMain. Executes all available reports for metadata QA
* Fixup, use ModelConverter.toModel to create Dataset[OreAggregation] in Reports
* Add centeralized loggoing to JSON-L and reports generation
* Fixup rightholder.name reporting
* Code cleanup


src/main/scala/dpla/ingestion3/confs/Ingestion3Conf.scala
```diff
@@ -96,6 +97,42 @@ class CmdArgs(arguments: Seq[String]) extends ScallopConf(arguments) {
...
+
+  /**
+    * Gets the provider short name from command line arguments
+    *
+    * @return Provider short name
+    */
+  def getProviderName() = providerName.toOption
+    .map(_.toString)
+    .getOrElse(throw new RuntimeException("No provider name specified."))
+
```
AudreyAltman: Good idea to make methods to get the variable definitions from the config file.

src/main/scala/dpla/ingestion3/utils/Utils.scala
```diff
@@ -147,7 +117,21 @@ object Utils {
...
+    * @param operation
+    * @param shortName
+    * @return
+    */
+  def createLogger(operation: String, shortName: String = ""): Logger = {
+    val logger: Logger = LogManager.getLogger("ingestion3")
+    val appender = Utils.getFileAppender(shortName, operation)
+    logger.addAppender(appender)
+    logger
+  }
```
AudreyAltman: `createLogger` 👍

src/main/scala/dpla/ingestion3/entries/ingest/EnrichEntry.scala
```diff
@@ -87,22 +102,22 @@ object EnrichEntry {
       ExpressionEncoder.tuple(RowEncoder(model.sparkSchema), ExpressionEncoder())
 
     // Load the mapped records
-    val mappedRows: DataFrame = spark.read.avro(inputDir)
+    val mappedRows: DataFrame = spark.read.avro(dataIn)
```
AudreyAltman: I like that you changed this variable name from `inputDir` to `dataIn`.

src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
```diff
@@ -0,0 +1,73 @@
...
+  * Single entry point to run harvest, mapping, enrichment, indexing jobs and all reports.
+  * Support running against an already harvested data set and running a fresh harvest
+  *
+  * Expects four parameters:
+  *   1)  --output  A base path to save the mapping, enrichment, json-l and report outputs
+  *   2)  --input   A path to the harvested data
+  *   3)  --conf    A path to the application configuration file
+  *   4)  --name    Provider short name
+  */
+object IngestRemap {
```
AudreyAltman: I find the name of this object a little confusing - `Ingest` or `IngestEntry` would be my instinct.  However, you're more familiar with the workflow than me, so I think you should choose whatever name you like.

src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    // Read in command line args.
+    val cmdArgs = new CmdArgs(args)
+
+    val baseDataOut = cmdArgs.getOutput()
+    val confFile = cmdArgs.getConfigFile()
+    val shortName = cmdArgs.getProviderName()
+
+    // Outputs
+    val harvestDataOut = cmdArgs.getInput()
```
AudreyAltman: Wouldn't the above get `getOutput` instead of `getInput`?

src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
```diff
@@ -0,0 +1,73 @@
...
+
+    // Read spark master property from conf, default to 'local[1]' if not set
+    val sparkMaster = conf.spark.sparkMaster.getOrElse("local[1]")
+
+    val sparkConf = new SparkConf()
+      .setAppName(s"Mapping: $shortName")
+      .set("spark.serializer", "org.apache.spark.serializer.KryoSerializer")
+      .set("spark.kryoserializer.buffer.max", "200")
+      .setMaster(sparkMaster)
+
```
AudreyAltman: Don't you need to execute the harvest before you execute the mapping?  Does that happen somewhere else?

src/main/scala/dpla/ingestion3/reports/Report.scala
```diff
@@ -29,6 +28,7 @@ import util.Try
   */
 trait Report {
 
+  private final val missing = "__MISSING__"
```
AudreyAltman: :+1: making `missing` a `private final val`

src/main/scala/dpla/ingestion3/utils/Utils.scala
```diff
@@ -63,7 +50,7 @@ object Utils {
     * @param runtime Runtime in milliseconds
     * @param recordCount Number of records output
     */
-  def logResults(runtime: Long, recordCount: Long, logger: Logger): Unit = {
+  def logResults(runtime: Long, recordCount: Long): String = {
```
AudreyAltman: Could we change the name of this method since it isn't actually logging anything anymore?

src/main/scala/dpla/ingestion3/entries/ingest/EnrichEntry.scala
moltude: Trying to normalize these var names across the project..

src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
moltude: Yeah, I went back and forth on this. This entry point is just supposed to handle testing a mapping change (e.g. testing something on CQQ). I thought about CqaReingest but nothing really sounded great. 

Open to other names..

src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
moltude: No, b/c this entry point expects harvest data already the "input" on the command line is the harvest data and the "output" is a base directory where we will put the output from mapping, enrichment, reports and jsonl.

base directory + "/mapping"
base directory + "/enrichment"
base directory + "/reports"
base directory + "/jsonl"

src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
moltude: Harvest needs to be run on its own.

src/main/scala/dpla/ingestion3/utils/Utils.scala
moltude: good call

moltude: Actually, I think I renamed this in the other massive PR...

https://github.com/dpla/ingestion3/blob/1918cf27001a1cc1a32e73db54bbff273a50710e/src/main/scala/dpla/ingestion3/utils/HttpUtils.scala

src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
AudreyAltman: Oh I see.  Why wouldn't you want this Object to harvest the data?  What's the motivation for keeping that part of the workflow separate?

src/main/scala/dpla/ingestion3/utils/Utils.scala
AudreyAltman: :+1: great minds... ;)

src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
moltude: My goal with this was to have an entry point that is going to allow us to plow through CQA processes as we start to migrate providers and since harvest is the most expensive operation I thought it made sense to omit it here. 

I did have an earlier version where running a harvest was optional but since the letter of the ticket was map, enrich, index, report I dropped it off.  There should be another entry point for production that omits the report generation so this shouldn't be seem as the only execution point for us. Just one designed for testing mappings.

AudreyAltman: That makes sense.  Thanks!

src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
mdellabitta: `WholeShebang`

src/main/scala/dpla/ingestion3/entries/ingest/JsonlEntry.scala
```diff
@@ -35,15 +39,34 @@ object JsonlEntry {
...
+    executeJsonL(sparkConf, inputName, outputName, Utils.createLogger("jsonl", ""))
+  }
+
+  /**
+    * Generate JSON-L files from AVRO file
+    * @param sparkConf Spark configuration
+    * @param dataIn Data to transform into JSON-L
+    * @param dataOut Location to save JSON-L
+    */
+  def executeJsonL(sparkConf: SparkConf, dataIn: String, dataOut: String, logger: Logger) = {
```
mdellabitta: I think eventually we should break out these `execute` methods to their own traits or classes and have the `Entries` just have the single main method. 

Then that trait or abstract class can have methods to inspect the results of the run and make decisions about it. Would also be useful for integrating into something like Combine.

src/main/scala/dpla/ingestion3/entries/ingest/JsonlEntry.scala
```diff
@@ -52,10 +75,9 @@ object JsonlEntry {
       }
     )
 
-    indexRecords.coalesce(1).write.text(outputName)
-
+    indexRecords.coalesce(1).write.json(dataOut)
```
mdellabitta: Unrelated question: Do we need this `coalesce` here? It probably slows things down quite a bit.

moltude: Not sure but I thought we needed it to get all of the output into a single file?

mdellabitta: I'm assuming in asking this that we don't need a single file. Do we?

moltude: I suppose if having multiple files isn't a negative then there is no reason to coalesce. But if we are indexing files by hand having to run elasticdump multiple times could be a real pain.

mdellabitta: Lookie here! https://www.npmjs.com/package/elasticdump#multielasticdump

PR #118, opened 2017-10-31
edmRights reports

moltude:
Adds edmRights to PropertyValue and DistinctPropertyValue reports.
Fixes up PA mapping of rights and edmRights to property test reports.


moltude:
bump for review

PR #121, opened 2017-11-07
Fixes to PA mapping

moltude:
Addressees DT-1833 
* Lowercase `type` values
* Get distinct values for `date`, `format`, `genre`, `language`, `place`, `publisher`
* Remove empty values for JSON-L export
* Fix provider name ("PA Digital")
* Add `edm:rights` to JSON-L export
* Remove print statement logging from HttpUtils


src/main/scala/dpla/ingestion3/model/package.scala
```diff
@@ -37,6 +39,8 @@ package object model {
 
   def eitherStringOrUri(uri: URI): LiteralOrUri = new Right(uri)
 
+  def dedup(seq: Seq[String]): Seq[String] = seq.distinct
```
mdellabitta: Looks like you don't need this...

moltude: oh yeah forgot to strip that out...

PR #125, opened 2017-12-14
Universal string and dedupe enrichments

AudreyAltman:
This adds standard enrichments for the following:
* Deduplicate repeated values in multi-value fields
* Replace double-whitespace with single-whitespace
* Remove HTML markup

This also introduces a different model for organizing enrichments that we could consider using for other enrichments if the team likes it.

Methods that enrich a single strings are now in `StringUtils` and organized into an implicit class.  In any domain in which `StringUtils` are imported, you can call them directly on a string, e.g. `"mystring".stripPunctuation.reduceWhitespace.stripHTML` etc.  I thought the ability to call a method directly on a string would make it easier to read and organize enrichments.  We could follow this format enrichments on other value types, not just string enrichments, if we like it.

`StringEnrichments` and `DeDuplicationEnrichment` each take an `OreAggregation`, apply enrichments, and return an `OreAggregation`.  In `StringEnrichments`, I did my best to select which fields it made sense to apply the enrichments to, considering that some of the fields are from MAP 4.1 and therefore not yet in use - feedback would be appreciated about whether I should include or exclude any fields.

`StandardEnrichmentUtils` collects all the standard enrichments in another implicit class.  I used an implicit class again b/c I thought it would make it easier to read and organize enrichments, especially if we need to apply many enrichments in a particular order.  The idea is that in any given domain, you could import whatever enrichment utils you needed, and call them like this `myOreAggregation.deDuplicate.standardStringEnrichments.anotherEnrichment` etc.

### DIFF
--- a/src/main/scala/dpla/ingestion3/enrichments/DeDuplicationEnrichment.scala
+++ b/src/main/scala/dpla/ingestion3/enrichments/DeDuplicationEnrichment.scala
@@ -1,0 +1,73 @@
+package dpla.ingestion3.enrichments
+
+import dpla.ingestion3.model._
+
+/**
+  * Enrichment to remove duplicate values in multi-value fields.
+  *
+  */
+class DeDuplicationEnrichment {
+
+  /**
+    * Main entry point.
+    *
+    * @param record OreAggregation
+    * @return enriched OreAggregation
+    */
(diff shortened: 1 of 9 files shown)


moltude:
👍

PR #126, opened 2017-12-15
Fix/mdl mapping

moltude:
* Mapping error no longer include full stacktrace
* A number of initial mapping fixes for MDL 
* extractStrings() in JsonExtractionUtils now returns unique values
* Add iterify() method to JsonExtractionUtils to ensure that all JValues are wrapped in a JArray
* Add collection.description and language.iso_639_3 properties to JSON-L export

### DIFF
--- a/src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
+++ b/src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
@@ -63,10 +63,10 @@ object IngestRemap {
 
     // Json-l
     logger.info(s"Saving JSON-L output to: $jsonlDataOut")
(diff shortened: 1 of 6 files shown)


mdellabitta:
To be honest, I'm a little concerned by "extractStrings() in JsonExtractionUtils now returns unique values" because it removes the possibility of getting two lists of strings and cross referencing them with one another. Some of the `distinct` stuff is being handled by Audrey's PR, is that a better place to handle that?

AudreyAltman:
FWIW the standard enrichment I wrote is intentionally very thorough, so it should be ok to bypass it here, unless there’s a specific reason we need it at mapping.

AudreyAltman:
I have no concerns to add to @mdellabitta’s so I will approve. Thanks @scott!

moltude:
I'll remove

PR #132, opened 2018-02-06
Add support for parsing yyyy-yyyy date intervals

moltude:


src/test/scala/dpla/ingestion3/enrichments/ParseDateEnrichmentTest.scala
```diff
@@ -131,6 +131,19 @@ class ParseDateEnrichmentTest extends FlatSpec
...
   }
 
+  it should "parse YYYY-YYYY date to iso date" in {
+    val date = "2001-2003"
+    val originalDate = EdmTimeSpan(originalSourceDate = Some(date))
+    val enrichedDate = EdmTimeSpan(
+      originalSourceDate = Some(date),
+      prefLabel = Some("2001-2003"),
+      begin = Some("2001-01-01"),
+      end = Some("2003-01-01")
```
mdellabitta: Dumb question: Do dates that are year only have a different value for month and day depending on whether they are begin or end? Does 2001-2003 mean Jan 1, 2001 to Jan 1, 2003, or Jan 1, 2001 to Dec 31, 2003?

moltude: Good question. I've only ever seen a year value evaluated to "Jan 1" so I'm not sure. I can certainly see the argument for it.

mdellabitta: Let's go with this then, rather than begging for complication!

PR #134, opened 2018-02-07
Ohio mapping fixes

moltude:
- Refactor of Ohio mapping and associated changes 
  - Cleanup flatMap(split).filter() calls into splitOnDelimiter()
  - Add capitalizeFirstChar() string enrichment. Capitializes the first alphanumeric character in a given string
  - Remove type mapping from OhioExtractor and use universal DCMIType enrichment 
- Reports for language now include SkosConcept.concept in addition to SkosConcept.providedLabel. We need to report on concept because that is where the enriched value is stored. 
- EnrichmentDriver now invokes the DCMIType enrichment which looks up the original type value against a mapping of terms to DCMIType IRIs. These IRIs are then used to lookup a label for the IRI and that string is finally downcased. 
- OaiResponseProcessor now ignores records marked as "deleted" in an OAI feed
- Change ElasticSearch JSON-L mapping to send the enriched value to the language.iso639 field.
    language.name will be either SkosConcept.concept (enriched value) and if that does not exist then map the original value from SkosConcept.providedLabel.

### DIFF
--- a/src/main/scala/dpla/ingestion3/enrichments/EnrichmentDriver.scala
+++ b/src/main/scala/dpla/ingestion3/enrichments/EnrichmentDriver.scala
@@ -43,7 +43,16 @@ class EnrichmentDriver(conf: i3Conf) extends Serializable {
       sourceResource = enriched.sourceResource.copy(
         date = enriched.sourceResource.date.map(d => dateEnrichment.parse(d)),
         language = enriched.sourceResource.language.map(l => LanguageMapper.mapLanguage(l)),
-        place = enriched.sourceResource.place.map(p => spatialEnrichment.enrich(p))
+        place = enriched.sourceResource.place.map(p => spatialEnrichment.enrich(p)),
+        `type` = enriched.sourceResource.`type`.map(t => {
+          // Cast original type value to lower case and map it to DCMIType IRIs (@see DcmiTypeMap).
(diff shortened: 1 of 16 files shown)


src/main/scala/dpla/ingestion3/enrichments/EnrichmentDriver.scala
```diff
@@ -43,7 +43,16 @@ class EnrichmentDriver(conf: i3Conf) extends Serializable {
       sourceResource = enriched.sourceResource.copy(
         date = enriched.sourceResource.date.map(d => dateEnrichment.parse(d)),
         language = enriched.sourceResource.language.map(l => LanguageMapper.mapLanguage(l)),
-        place = enriched.sourceResource.place.map(p => spatialEnrichment.enrich(p))
+        place = enriched.sourceResource.place.map(p => spatialEnrichment.enrich(p)),
+        `type` = enriched.sourceResource.`type`.map(t => {
```
mdellabitta: This might be cleaner as a chain of map calls:
```

`type` = enriched.sourceResource.`type`.map(_.toLowerCase)
.map(DcmiTypeMapper.mapDcmiType)
.flatMap(DcmiTypeStringMapper.mapDcmiTypeString)
.map(_.toLowerCase)
```

moltude: much clearer.

PR #136, opened 2018-02-21
Language enrichment multi-file support

moltude:
Adds support in the language enrichment for loading code to term mappings from multiple files. 
Adds iso639-2/3 and DPLA-specific code mappings to project


src/main/scala/dpla/ingestion3/enrichments/VocabEnforcer.scala
```diff
@@ -140,26 +139,50 @@ object DcmiTypeStringMapper extends VocabEnforcer[String] {
...
 
 /**
-  * Reads in ISO-693-3 data from tab-delimited file and generates a
+  * Reads in iso693-X data from CSV files and generates a
   * map to reconcile language abbreviations to the full term
   *
   */
 object LanguageMapper extends VocabEnforcer[String] {
+
+  private val languageFiles = Seq(
```
mdellabitta: This will not work if the project is ever bundled into a JAR. Basically everything in src/main/resources gets bundled into the top level directory of the JAR. 

The better way to do this is to use the classpath/classloader to locate the file. Unfortunately I don't know off the top of my head how to do that idiomatically in Scala (I know the Java version).

src/main/scala/dpla/ingestion3/enrichments/VocabEnforcer.scala
```diff
@@ -140,26 +139,50 @@ object DcmiTypeStringMapper extends VocabEnforcer[String] {
...
+    * and the original value in SkosConcept.providedLabel
+    */
+  val mapLanguage: (SkosConcept) => SkosConcept = matchToSkosVocab(_, languageMap)
+
+
+  /**
+    * Gets an ordered version of the language map
+    * @return
+    */
+  def getLanuageMap = ListMap(languageMap.toSeq.sortBy(_._1):_*)
```
mdellabitta: Dumb question: Why do you need this to be sorted?

src/main/scala/dpla/ingestion3/enrichments/VocabEnforcer.scala
```diff
@@ -140,26 +139,50 @@ object DcmiTypeStringMapper extends VocabEnforcer[String] {
...
-      .map(f => f(0) -> f(1))
-      .toMap
-  }
-  // Attempt to enrich the original record value
-  val mapLanguage: (SkosConcept) => SkosConcept = matchToSkosVocab(_, iso639Map)
+  private val languageMap  =
+    /**
+      * Spark could be used with pattern matching to load CSV for TAB files in a dir
+      * but that seems like overkill here. Probably the best way to go for now is supporting
+      * passing a list of file paths to load and then we can inject those values via a config
```
mdellabitta: Maybe a subdirectory under `resources`. Then you don't have to worry about config.

src/main/scala/dpla/ingestion3/enrichments/VocabEnforcer.scala
moltude: Yeah I fought with that for about two hours this morning before giving up and just trying to get done. Insanely frustrating.

src/main/scala/dpla/ingestion3/enrichments/VocabEnforcer.scala
moltude: There is a comment in the Reports file about exporting the map of values we used to enrich the data and this would be a part of that report. I was also dumping the data out to check that that the files were getting merged correctly. So this method doesn't have immediate use but will

src/main/scala/dpla/ingestion3/enrichments/VocabEnforcer.scala
moltude: I couldn't get this to work with files directly under `resources` or nested in folders under `resources`. Agree this needs a fix

PR #137, opened 2018-02-21
DO NOT MERGE: Mapping refactor

moltude:
Refactor of mapping model and all provider implementations. This is the prerequisite to doing the actual "collect all errors" work.


src/test/scala/dpla/ingestion3/mappers/providers/NaraExtractorTest.scala
```diff
@@ -47,17 +41,24 @@ class NaraExtractorTest extends FlatSpec with BeforeAndAfter {
 
   it should "extract collections" in {
     val collections = extractor.collection(xml)
-    assert(collections === Seq("Records of the Forest Service"))
+    // According to the mapping this should belong to three collections:
```
moltude: @mdellabitta We should probably checkin about all these NARA changes. I am really unsure how I managed to completely break and fix the NARA test suite with these refactoring changes. I also don't understand how some of these tests were passing with the test data.

moltude: I think I've accidentally redone quite a bit of the work on this branch: https://github.com/dpla/ingestion3/tree/feature/dt-1496-nara-bugfixes.  What is that status of this branch? If the changes on that branch are still valid I can try to move them over to this branch to pickup on this PR or just leave it alone for now and pick back up after this PR is merged in.

moltude:
DO NOT MERGE.

PR #138, opened 2018-02-22
Mapping refactor

moltude:
This is a major refactor of the flow and control of mapping as well as the provider implementations.

A provider mapping is responsible for implementing all of the field mappings
defined in the Mapping[T] trait. These provider mappings also extend typed
Extraction and IdMinting traits (XmlExtractor and JsonExtractor respectively).
This allows us to test field mappings individually rather than evaluating the
entire mapped record result.

This PR also includes a number of naming changes. Provider mapping classes are
now called Mappings (previously Extractors). The Mapper trait and class
implementations (XmlMapper and JsonMapper) define and implement the mapping of an
entire record. Within the #map() method it calls each of the field mapping
functions defined the relevant Mapping class).

These changes make the provider Mappings easier to test and they are the
prerequisite for collecting all mapping errors from a record.

Providers have profiles in code where they specify the type of record (JSON/XML)
as well as the corresponding components of the stack (harvester, mapping and
possibly enrichment profiles (TBD))

This PR also consolidates some files in the mappers/utils package
* JsonExtractionUtils.scala
* XmlExtractionUtils.scala

are all in the mapping/utils/Extractor.scala


src/test/scala/dpla/ingestion3/mappers/providers/NaraMappingTest.scala
```diff
@@ -3,27 +3,27 @@ package dpla.ingestion3.mappers.providers
...
-  val extractor = new NaraExtractor(xmlString, shortName)
+  val extractor = new NaraMapping
 
-  "A NaraExtractor" should "successfully extract from a valid document" in {
-    extractor.build() match {
-      case Success(data) => succeed
-      case Failure(exception) => fail(exception)
-    }
-  }
+//  "A NaraExtractor" should "successfully extract from a valid document" in {
```
mdellabitta: Probably should just nuke these two commented out tests

moltude: 💥

PR #140, opened 2018-02-23
Add Ohio tests

moltude:
* Add sample XML data for Ohio (ohio.xml) 
* Add test suite for Ohio mapping 
* Minor mapping fixes for Ohio

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/OhioMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/OhioMapping.scala
@@ -88,7 +88,7 @@ class OhioMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeS
         case "dc" => r.text
         case _ => ""
       }
-    }) // FIXME This shouldn't be returning Seq("") if no dc:rights exist
+    }).filter(_.nonEmpty)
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/OhioMapping.scala
```diff
@@ -87,7 +88,7 @@ class OhioMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeS
         case "dc" => r.text
         case _ => ""
       }
-    }).filter(_.isEmpty)
+    }) // FIXME This shouldn't be returning Seq("") if no dc:rights exist
```
mdellabitta: Do you want to handle this now?

src/test/resources/ohio.xml
```diff
@@ -0,0 +1,86 @@
+<record>
```
mdellabitta: I think in general, the inline xml way of writing tests will be more useful, because many mappings have branches in them depending on what the document has, and you can't test that situation with a single example file. 

If that situation arises with Ohio, you're going to end up with a hybrid of test data. Just thinking it might make sense to start with the inline xml approach.

PR #141, opened 2018-03-01
Type enrichment case match failure bug

moltude:
This addresses a bug where original values that could not be mapped to a DCMIType IRI would return None and fail the match. Implements previous suggestion by @mdellabitta


moltude:
Forgot to add that this also removes the spatial enrichment because of ongoing issues the the quality of Twofishes responses.

mdellabitta:
Can we talk about that? We're using that in production, so if it's really so bad that we need to just remove it here, we have a bigger problem.

moltude:
@AudreyAltman bumping

PR #144, opened 2018-03-07
stripBrackets enrichment

moltude:
[Addresses IN-350](https://digitalpubliclibraryofamerica.atlassian.net/secure/RapidBoard.jspa?rapidView=9&projectKey=IN&selectedIssue=IN-350)

Adds string normalization member `stringBrackets`. This method removes a defined set of paired characters (e.g {}, [], ()) from a provided string. It constructs two regular expressions, one to search the string for a match and a second to identify characters to replace. Both of these patterns are built from a base pattern (see `baseBracketSearchPattern` and `baseBracketReplacePattern`) and the target characters passed in (see `brackets` member and `Bracket` case class). Both of the regex patterns ignore and replace extraneous white space respectively.

### DIFF
--- a/src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
+++ b/src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
@@ -129,6 +129,7 @@ object StringUtils {
         value
     }
 
+
     /**
       * Replaces the character at the specified index in s with c
       * @param s String value
@@ -142,7 +143,7 @@ object StringUtils {
       * Iterates over the string to find the first alphanumeric char
       * and returns the index
       *
-      * @param str
+      * @param str String to search
       * @return
       */
     private def findFirstChar(str: String): Int = {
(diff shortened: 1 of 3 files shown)


src/test/scala/dpla/ingestion3/enrichments/StringUtilsTest.scala
```diff
@@ -198,4 +198,41 @@ class StringUtilsTest extends FlatSpec with BeforeAndAfter {
     val enrichedValue = originalValue.findAndRemoveAll(stopWords)
     assert(enrichedValue === "photograph")
   }
+
+  "stripBrackets" should "remove leading and trailing ( )" in {
```
moltude: @guegueng Can you look at these tests for `stripBrackets` and let me know if you think I've addressed all the possible cases here?

src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
```diff
@@ -161,5 +162,56 @@ object StringUtils {
       *   whitespace (\s)
       */
     private def beginAndEndPunctuationToRemove = """[;:/,-\\t\\r\\n\s]"""
+
+    /**
+      * Removes matching leading and trailing brackets (square, round and curly braces)
+      * from a string
+      */
+    val stripBrackets: SingleStringEnrichment = {
+      val replacemnetRegex = for( (p, r) <- bracketPatterns if value.matches(p)) yield r
```
AudreyAltman: I think you meant to spell this `replacementRegex`.  Looks like the mis-spelling is consistent and not causing any bugs, so no biggie.

src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
```diff
@@ -161,5 +162,56 @@ object StringUtils {
...
+      // TODO is there a cleaner or safer way to convert Iterator[String] to String than mkString?
```
AudreyAltman: FWIW I do not know of another way to do the conversion.

moltude: thanks --0 I'll fix it when I rebase

PR #145, opened 2018-03-07
Strip ending period

moltude:
[Addresses IN-351](https://digitalpubliclibraryofamerica.atlassian.net/browse/IN-351)

Add `stripEndingPeriod` member to `StringUtils`
- Removes singular trailing period from a string
- Ignores and removes trailing white space
- Ignores leading, interior and multiple trailing periods (i.e. ellipsis)

Fixup CDL subject mapping
- Call `stripEndinPeriod`
- Call `cleanupLeadingPunctuation` and `cleanupTrailingPunctuation`

### DIFF
--- a/src/test/scala/dpla/ingestion3/enrichments/StringUtilsTest.scala
+++ b/src/test/scala/dpla/ingestion3/enrichments/StringUtilsTest.scala
@@ -210,7 +210,7 @@ class StringUtilsTest extends FlatSpec with BeforeAndAfter {
     val enrichedValue = originalValue.findAndRemoveAll(stopWords)
     assert(enrichedValue === "photograph")
   }
-
+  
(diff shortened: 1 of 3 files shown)


src/test/scala/dpla/ingestion3/enrichments/StringUtilsTest.scala
```diff
@@ -198,4 +198,35 @@ class StringUtilsTest extends FlatSpec with BeforeAndAfter {
     val enrichedValue = originalValue.findAndRemoveAll(stopWords)
     assert(enrichedValue === "photograph")
   }
+
```
moltude: @guegueng Can you look at these tests and let me know if you see any obvious cases not addressed?

PR #148, opened 2018-03-07
IMT, MIME, content-type format filter

moltude:
[Addresses IN-356](https://digitalpubliclibraryofamerica.atlassian.net/browse/IN-356)
This is a refactor of the format filter implemented for the Ohio mapping

The format filter now works more like the vocab enrichment in that it reads term lists from specified files. Generates a huge regular expression from those terms and uses that regex to search and remove invalid format terms.

- Format files are under /resources/formats/
- Refactor Ohio format filter invocation in OhioMapper
- Add additional tests


moltude:
@guegueng As always if you can take a quick look at the `StringUtilsTest` file and see if there are any obvious cases that are not being met by those tests.

src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
```diff
@@ -226,6 +214,66 @@ object StringUtils {
...
+      // in the stop words list
+      val stopWords = getFormatStopwords
+        .map(
+          // TODO is there a better way to 'auto-escape' reserved regex chars?
+          _.replace("""/""", """\/""")
+          .replace("""+""", """\+""")
+          .replace("""-""", """\-"""))
+
+      // Create on big'ol regex from those stop words that ignores
+      // case and white space.
```
moltude: This comment isn't 100% correct as the white space is still important to avoid premature replacements (e.g. the value is `JPEG 2000 photograph` it was matching JPEG first and returning `2000 photograph`

src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
```diff
@@ -266,8 +266,13 @@ object StringUtils {
           .replace("""+""", """\+""")
           .replace("""-""", """\-"""))
 
-      // Create on big'ol regex from those stop words that ignores
-      // case and white space.
+
+      // FIXME There is a case not easily addressed here where a stop word term exists inside a larger term.
```
moltude: @guegueng calling this out and a possible discussion topic for us next Friday.

src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
```diff
@@ -226,6 +214,66 @@ object StringUtils {
...
+
+      // Create on big'ol regex from those stop words that ignores
+      // case and white space.
+      val regex = "(?i)" + stopWords
+        .map(word => s"($word)")
+        .mkString("|")
+
+      // Perform the removal and trim white space
+      value.replaceAll(regex, "").trim
+    }
```
AudreyAltman: Have you considered that instead of creating one huge regex, you could iterate through the stop words, replacing them one at a time?  That way if there was a problem or error with one of the stop words, the others would still get cleaned up.  Not sure what its impact would be on efficiency, though.  Just a thought, feel free to keep one-regex-to-rule-them-all if you think it best.

src/test/scala/dpla/ingestion3/enrichments/StringUtilsTest.scala
```diff
@@ -204,13 +204,42 @@ class StringUtilsTest extends FlatSpec with BeforeAndAfter {
...
+    val originalValue = "3gpdash-qoe-report+xml photograph  "
+    val enrichedValue = originalValue.stripInvalidFormats
+    assert(enrichedValue === "photograph")
+  }
+  it should "remove a stop word if it exists within another term" in {
+    val originalValue = "application/xmlphotograph"
+    val enrichedValue = originalValue.stripInvalidFormats
+    assert(enrichedValue === "photograph")
+  }
+
```
AudreyAltman: Just a quick sanity check - is it okay that some of the tests return `Photograph` (uppercase `P`) and others return `photograph` (lowercase `p`)?  I assume any upcasing or downcasing enrichments are handled elsewhere?

src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
moltude: That was kind of what I was doing before but I found more flexibility and success with regex recently but I agree that this may not be the most sustainable.

src/test/scala/dpla/ingestion3/enrichments/StringUtilsTest.scala
moltude: that is a sanity check that the enrichment isn't altering the case of the original record even though the matching process is case insensitive. This was something that is much cleaner w/regex

src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
AudreyAltman: I trust your experience and judgement - you have done a lot more with enrichments than I!

moltude:
@AudreyAltman I've made some additional changes to this PR since you reviewed it but I don't think it changes things too much. Just tweaked the regex and hopefully made things cleaner and addressed some additional edge cases (stop word inside of term and cleanly dealing with stop words that contain white spaces).

PR #149, opened 2018-03-08
Ohio mapping and test fixes

moltude:
Fixes after reviewing [Ohio mapping, specific mapping enrichments, and universal enrichments](https://docs.google.com-/spreadsheets/d/1nUQmlW959p3j9sR-ic5cYL66mGUSIoIuSD550Ic1hk4/edit#gid=0)

**Ohio mapping changes**
- `collection` mapping maps all occurrences of `isPartOf`
- `creator` mapping splits original values on `;`
- `contributor` mapping splits values on `;`
- Remove unnecessary `capitalizeFirstChar` call in `format` mapping (this is applied in the universal enrichment for `format`). 
- Refactor `dataProvider` mapping to make it cleaner
- Refactor `isShownAt` so that it maps only the first occurrence of the `isShownAt` property if more than one is provided. 
- Add additional tests  

**Universal enrichment changes**
- SkosConcept.title
  - Add `stripBrackets`
- SkosConcept.concept
  - Add `stripBrackets`
  - Add `stripEndingPeriod`
  - Add `capitalizeFirstChar`
- SkosConcept.providedLabel
  - Add `stripBrackets`
  - Add `stripEndingPeriod`
  - Add `capitalizeFirstChar`
- EnrichmentDriverTest -- `eng` changed to `Eng` because of the addition of `capitalizeFirstChar` normalization to `SkosConcept.concept` property

**Performance**
Ohio mapping runtime increased to 20+ minutes using the original implementation of `stripInvalidFormats`. The original relied on building a single regex that included all 1,300+ stop words. Compiling a list of regular expressions to apply individually and iterating through them got the runtime down to four minutes but that was still double the original runtime. The bulk of the IANA terms were removed because they appeared unlikely to appear in our data. Reducing the terms list brought the runtime back down to two minutes
- Reduced IANA stop words list from ~1,200 terms to ~130
- Refactor stripInvalidFormat to create list of regular expressions rather than one large one
- Apply regex using `foldLeft`. 

Code cleanup and formatting


src/main/scala/dpla/ingestion3/enrichments/StringUtils.scala
```diff
@@ -16,6 +16,9 @@ import scala.util.matching.Regex
 object StringUtils {
   implicit class Enrichments(value: String) {
 
+    // TODO Move this out of the method so the file aren't re-read for every record (see VocabEnforcer also)
```
mdellabitta: Ooof, yeah, that's going to kill performance.

mdellabitta: What if you just bumped it up to being a member of StringUtils?

moltude: That does improve the performance after adding back the 1,200 terms (2:30 runtime) but made negligible difference when running with the reduced terms list (1:50 and 2:00 runtimes). I think testing over a large record set than Ohio (95,000 records) would be the next step.

I see two problems here
- Inefficiently loading and reloading the data from files (which I think your suggestion addresses) 
- Searching for a lot of unnecessary IMT/content-type strings in the original record. 

The second is clearly much easier to solve in the short-term by just deleting terms and just doing that got performance back to where it was before even with re-loading the truncated files for every record but that approach leaves us vulnerable in the future if our terms list does grow substantially.

I'll commit the change moving the term loading functionality out of the implicit class but should I also add back all those 1,200 terms or do we think that the best approach is a combination of the refactor and a truncated IANA term list?  (@guegueng can you weigh in on the terms list as well?)

mdellabitta: Correct me if I'm wrong, but those are in a List or other type of Seq? Making them keys in a Map or using a Set will probably improve lookup performance O(N) -> O(1)

moltude: Assuming you are asking about the regular expressions constructed from the Seq[String] of stop words read from the file? In that case [regex here](https://github.com/dpla/ingestion3/pull/149/files#diff-445e7336e847f8ce1cde56db71f716b1R274) is already a map. 

`foldLeft` might not be the way to do it or I might not fully understand whats up here

mdellabitta: The stopwords... Generally speaking, if you have to loop over a whole list, it's going to get slower as the list gets larger. But Sets and Maps are generally optimized for quick lookup.

mdellabitta: Just to be clear, i'm not talking about the method `map`, but the data structure that contains Stopwords.

moltude: Got it. I think these changes address that issue.

PR #150, opened 2018-03-19
Feature blockFilter and allowFilter

moltude:
Adds support for filtering allowed and blocked terms from a string. These fuctions are requirements for the Format and Type normalization and enrichment operations. 

See [MAPv5 enrichment descriptions](https://docs.google.com/spreadsheets/d/1nUQmlW959p3j9sR-ic5cYL66mGUSIoIuSD550Ic1hk4/edit#gid=0)

### `FilterList` ###

To be implemented when defining set of terms that are prohibited or allowed in a string. Terms can be defined in code or via files path(s) to be read. 

**Hard coded term list**
```
object FormatTypeValuesBlockList extends FilterList {
  override val termList: Set[String] = Set("Image")
}
```

**Read terms from files**
```
object BadFormatsList extends FilterList {
  val termList: Set[String] = getTermsFromFiles
  override val files: Seq[String] = Seq("/formats/bad_formats.csv")
}
```

### `FilterRegex` ###

Base regular expression patterns to be combined with the `FilterList.termList` to create appropriate regular expressions for use with the `applyAllowFilter()` and `applyBlockFilter()`.  Additional base formulas can be added if fields or partners have specific requirements. 

* `allowListRegex` - A case-insensitive whole term match regex that ignores leading and trailing white space. This pattern matches the blockListRegex because of the simplicity of these patterns. However, if they become more complicated to support removing terms from within a string or ignoring punctuation or term sepeators then a combination of regular expressions may be required when using `applyAllowFilter()`
* `blockListRegex` - A case-insensitive whole term match regex that ignores leading and trailing white space

### `StringUtils` ###

* `applyAllowFilter()` Accepts a set of Strings to preserve in the original value. Those string values can be literal strings or regular expressions. *Construction and testing of those regular expressions is dependent on the user.* Regular expressions can allow for filtering out terms while:
  * ignoring white space
  * ignoring case
  * ignoring punctuation
  * greedy matching on plurals

* `applyBlockFilter()` Applies the provided set of literals or patterns and removes all matches from the original string

**Term files** 
* `resources/types/dpla_type_normalization.csv` - Pairs local type values to a DCMI type labal (e.g. `drawing -> image`). Used to filter non-matching values out of the `type` field as well normalizing local terms to a standard set of DCMI type values.

**Misc**
* Code documentation cleanup
* Add tests and fixup existing Ohio tests


src/main/scala/dpla/ingestion3/enrichments/FilterEnrichment.scala
```diff
@@ -0,0 +1,100 @@
...
+  *     override val termList = Set[...].map(_.blockListRegex)
+  *   }
+  *
+  */
+trait FilterList {
+  // Set of terms to filter around (either block list or allow list)
+  val termList: Set[String]
+
+  // File paths to source termList from
+  val files: Seq[String] = Seq("")
```
mdellabitta: You probably just want Seq() here, right?

PR #152, opened 2018-03-26
Fixup PA thumbnail and intermediateProvider

moltude:
Add back mapping for thumbnail and intermediateProvider lost during mapping refactor.


src/main/scala/dpla/ingestion3/mappers/providers/PaMapping.scala
```diff
@@ -116,4 +121,21 @@ class PaMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq
...
+    * @param data
+    * @return
+    */
+
+  def thumbnail(implicit data: Document[NodeSeq]): ZeroToOne[EdmWebResource] = {
+    val ids = extractStrings(data \ "metadata" \\ "identifier")
+    if (ids.size > 2)
+      Option(uriOnlyWebResource(new URI(ids.last)))
+    else
+    None
```
AudreyAltman: Add an indentation to line 139 so it's the same as line 137?

PR #153, opened 2018-03-27
Update .gitignore

moltude:
Add src/main/resources/target/* to .gitignore


moltude:
I have no idea but I just deleted it.. this can be closed and deleted I suppose

PR #155, opened 2018-04-05
Add District Digital mapping

moltude:
Initial mapping and tests for Digital Digital (Washington D.C.). This mapping is very close to Ohio so `ohio.xml` sample data was reused for the D.C. tests.

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/DcMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/DcMapping.scala
@@ -1,0 +1,124 @@
+package dpla.ingestion3.mappers.providers
+
+import java.net.URI
+
+import dpla.ingestion3.enrichments.normalizations.StringNormalizationUtils._
+import dpla.ingestion3.mappers.utils.{Document, IdMinter, Mapping, XmlExtractor}
+import dpla.ingestion3.model.DplaMapData.{ExactlyOne, ZeroToOne}
+import dpla.ingestion3.model._
+import dpla.ingestion3.utils.Utils
+import org.json4s.JValue
+import org.json4s.JsonDSL._
+
+import scala.xml._
+
+
+class DcMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq] {
+  // ID minting functions
+  override def useProviderName(): Boolean = false // TODO confirm prefix
+
+  override def getProviderName(): String = "dc" // TODO confirm prefix
+
+  override def getProviderId(implicit data: Document[NodeSeq]): String = // TODO confirm w/gretchen
+    extractString(data \ "header" \ "identifier")
+      .getOrElse(throw new RuntimeException(s"No ID for record $data")
+      )
+
+  // SourceResource mapping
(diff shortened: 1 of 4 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/DcMapping.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  override def description(data: Document[NodeSeq]): Seq[String] =
+    extractStrings(data \ "metadata" \\ "description") ++
+      extractStrings(data \ "metadata" \\ "source")
+
+  override def format(data: Document[NodeSeq]): Seq[String] =
+    (extractStrings(data \ "metadata" \\ "format") ++
+      extractStrings(data \ "metadata" \\ "type"))
+      .map(_.applyBlockFilter(
+          DigitalSurrogateBlockList.termList ++
```
mdellabitta: I'd stick this concatenated list in a field somewhere rather than rebuilding it for every record.

src/main/scala/dpla/ingestion3/mappers/providers/DcMapping.scala
```diff
@@ -0,0 +1,138 @@
...
+import dpla.ingestion3.utils.Utils
+import org.json4s.JValue
+import org.json4s.JsonDSL._
+
+import scala.xml._
+
+
+class DcMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq] {
+  // ID minting functions
+  override def useProviderName(): Boolean = false // TODO confirm prefix
```
mdellabitta: Do these TODOs need addressing?

src/main/scala/dpla/ingestion3/mappers/providers/DcMapping.scala
```diff
@@ -0,0 +1,138 @@
...
+      case _ => None
+    }).headOption
+  }
+
+  override def isShownAt(data: Document[NodeSeq]): EdmWebResource =
+    uriOnlyWebResource(
+      Utils.createUri(extractStrings(data \ "metadata" \\ "isShownAt")
+        .headOption
+        .getOrElse(
+          throw new RuntimeException(s"No isShownAt property in record ${getProviderId(data)}")
```
mdellabitta: Might make sense to have a single, purpose-built exception we always throw for mapping errors.

moltude: Agreee

src/main/scala/dpla/ingestion3/mappers/providers/DcMapping.scala
moltude: good tip

src/main/scala/dpla/ingestion3/mappers/providers/DcMapping.scala
moltude: I'm not sure if we settled on using prefixes for all providers going forward as the default base DPLA id.  I think this can be addressed after initial CQA review

PR #156, opened 2018-04-11
LC Harvester

moltude:
Initial working version of the LC API harvester.

Gathers all item URLs to be fetched by iterating over the `sitemap` output of collection URLs. Distributes those item requests across all workers.  Swallows all errors and reports them out to `loc-harvest-xxx.log`. 

Observed throughput of this harvester is ~5 records per second or ~6 hours to harvest 150,000 records.

ingestion3.conf properties for LC 
```
# Library of Congress
loc.provider = "Library of Congress"
loc.harvest.type = "api"
loc.harvest.endpoint="http://www.loc.gov/collections/"
loc.harvest.rows=50
loc.harvest.setlist="american-revolutionary-war-maps,bain,civil-war-maps,detroit-publishing-company,fsa-owi-color-photographs,harris-ewing,national-photo-company,panoramic-maps"
```

### DIFF
--- a/src/main/scala/dpla/ingestion3/profiles/ProviderProfiles.scala
+++ b/src/main/scala/dpla/ingestion3/profiles/ProviderProfiles.scala
@@ -17,7 +17,6 @@ class CdlProfile extends JsonProfile {
 }
 
 /**
-<<<<<<< HEAD
   * District Digital
   */
 class DcProfile extends XmlProfile {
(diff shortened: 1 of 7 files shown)


src/main/scala/dpla/ingestion3/profiles/ProviderProfiles.scala
```diff
@@ -17,14 +17,24 @@ class CdlProfile extends JsonProfile {
 }
 
 /**
+<<<<<<< HEAD
```
mdellabitta: https://www.youtube.com/watch?v=IqycJpRdVaY

PR #158, opened 2018-04-19
LC mapping

moltude:
Intital mapping for Library of Congess.

LC Mapping 
* Update `ProviderRegistry.scala` with LC mapping values
* Add `LcMapping.scala`
* Add `LcMappingTest.scala`
* Add `lc.json` test data 

Other changes
* Comment out `date` and `place` enrichments from `EnrichmentDriver`. 
* Add `extractKeys` method to JsonExtractor to extract key values as Seq[String] from a given JObject. 
   * ex. ```{ "a": "b", "c": "d"}```  >>> Seq("a", "c")

### DIFF
--- a/src/main/scala/dpla/ingestion3/enrichments/EnrichmentDriver.scala
+++ b/src/main/scala/dpla/ingestion3/enrichments/EnrichmentDriver.scala
@@ -42,10 +42,10 @@ class EnrichmentDriver(conf: i3Conf) extends Serializable {
 
     enriched.copy(
       sourceResource = enriched.sourceResource.copy(
-        date = enriched.sourceResource.date.map(d => dateEnrichment.parse(d)),
+        // date = enriched.sourceResource.date.map(d => dateEnrichment.parse(d)),
         language = enriched.sourceResource.language.map(languageEnrichment.enrichLanguage),
-        `type` = enriched.sourceResource.`type`.flatMap(typeEnrichment.enrich),
-        place = enriched.sourceResource.place.map(p => spatialEnrichment.enrich(p))
+        `type` = enriched.sourceResource.`type`.flatMap(typeEnrichment.enrich)
+        //, place = enriched.sourceResource.place.map(p => spatialEnrichment.enrich(p))
       )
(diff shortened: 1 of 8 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/LcMapping.scala
```diff
@@ -0,0 +1,165 @@
...
+
+  // ID minting functions
+  override def useProviderName: Boolean = true
+
+  // Hard coded to prevent accidental changes to base ID
+  override def getProviderName: String =
+    "loc"
+
+  override def getProviderId(implicit data: Document[JValue]): String =
+    extractString(data.get \ "item" \ "id") // TODO confirm basis field for DPLA ID
```
moltude: @guegueng The ID minted in heidrun came from `library_of_congress_control_number` field whereas i3 is using `id`. The values overlap but I want to confirm that the control number is preferred for the sake of continuity between heidrun and i3. 

`library_of_congress_control_number = 73691632`
`id = http://www.loc.gov/item/73691632/`

https://github.com/dpla/heidrun/commit/6ab3f38f62156682a04579af8a98191bd072a05f#diff-1a4317d11c81f5bff6982f05e748c4bfR105

moltude:
This is ready fore review. We have a single outstanding issue related to image thumbnails but that may take some negotiation to resolve with partners.

src/main/scala/dpla/ingestion3/enrichments/EnrichmentDriver.scala
```diff
@@ -42,10 +42,10 @@ class EnrichmentDriver(conf: i3Conf) extends Serializable {
 
     enriched.copy(
       sourceResource = enriched.sourceResource.copy(
-        date = enriched.sourceResource.date.map(d => dateEnrichment.parse(d)),
+        // date = enriched.sourceResource.date.map(d => dateEnrichment.parse(d)),
```
moltude: We had discussed disabling these enrichments for the time being.

PR #159, opened 2018-04-19
Initial CO-WY file harvester

moltude:
Proof of concept -- based off of the NARA file harvester.  Still requries some cleanup and better integration into the gernalized harvester invocation (`HarvesterEntry`, HarvestExecutor`)


moltude:
@mdellabitta I'm attempting a refactor of this to integerate it into the larger harvester framework (single invocation, shared config etc.) but ran into a weird bug where the steam closes(?) after the first file. The single file invocation works but I'm having difficulity tracing the point of failure in the refactor.

PR #160, opened 2018-04-22
DO NOT MERGE: Harvester refactor

moltude:
Requires PR #159 

**Review only** 

Integrates P2P and Nara FileHarvesters with ApiHarvesters. Seperates out some functionality into new traits and tries to cut down on duplicate code but more work required.


src/main/scala/dpla/ingestion3/profiles/ProviderProfiles.scala
```diff
@@ -66,6 +66,16 @@ class OhioProfile extends XmlProfile {
   override def getMapping = new OhioMapping
 }
 
+/**
+  * P2P (Plains2Peaks)
+  * - Colorado and Wyoming
+  */
+class P2PProfile extends JsonProfile {
+  type Mapping = CdlMapping // FIXME Placeholder
```
mdellabitta: I'd use ??? for this. It has the nice property of blowing up when called.

mdellabitta: I should format that so it doesn't look like I left a placeholder in my comment: `???` I mean literally three question marks.

PR #161, opened 2018-04-23
TypeEnrichment mapping update

moltude:
Add new type mapping terms from OH and DC and fixup existing type mappings


src/main/scala/dpla/ingestion3/enrichments/TypeEnrichment.scala
```diff
@@ -15,57 +15,108 @@ class TypeEnrichment extends VocabEnrichment[String] {
...
     "appliance" -> dcmiType.Image,
+    "administrative record" -> dcmiType.Text,
+    "appliance" -> dcmiType.PhysicalObject,
+    "architectural drawing" -> dcmiType.Image,
+    "art print" -> dcmiType.Image,
     "audio" -> dcmiType.Sound,
+    "audiofile" -> dcmiType.Sound,
+    "black and white postcard" -> dcmiType.Image,
     "book" -> dcmiType.Text,
+    "bookchapter" -> dcmiType.Image,
```
AudreyAltman: Sanity check: "bookchapter" is `Image` and not `Text`?

moltude: good catch. @guegueng? I don't see that content type in the Type/Format doc table

guegueng: Bookchapter should be text. thanks for catching. This was a new one in PA's data, I didn't update the table, but now I realize I should do that. sorry about confusion there.

PR #176, opened 2018-06-18
Map logging refactor

moltude:
Initial work to collect error and warning information via a message system rather than raising Exceptions. 
* Add `messages` to `OreAggregation` data model so that all messages will travel with a record after it is mapped. Additional code required to support this change in `ModelConverter`
* Add cleaner reporting summary message (not fully utilized because of how message errors are dealt with)
* In `Mapper.map()` messages of level "ERROR" will raise an Exception. This is the key bridge abstraction as we migrate from Exceptions to IngestMessages.


moltude:
Those codacy complaints are illegitimate. :|

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -2,9 +2,9 @@ package dpla.ingestion3.model
 
 import java.net.URI
 
+import dpla.ingestion3.messages.IngestMessage
 import dpla.ingestion3.model.DplaMapData._
-import org.json4s.JValue
-import org.json4s.JNothing
+import org.json4s.{JNothing, JValue}
```
moltude: wth @codacy-badger?

src/main/resources/avro/MAPRecord.avsc
```diff
@@ -1079,6 +1079,36 @@
...
+      "name": "MessagesIngestMessageArray",
+      "type": "array",
+      "items": {
+      "name": "MessagesIngestMessage",
+      "type": "record",
+      "fields": [ {
+      "name": "message",
+      "type": "string"
+    }, {
+      "name": "level",
```
mdellabitta: Avro has Enums which you might want to consider. Will save space and constrain the level to a specific range of values.

src/main/scala/dpla/ingestion3/enrichments/normalizations/StringNormalizationUtils.scala
```diff
@@ -53,7 +54,10 @@ object StringNormalizationUtils {
       */
     lazy val applyBlockFilter: Set[String] => String = (termList) => termList
       .foldLeft(value) {
-        case (string, pattern) => string.replaceAll(pattern, "").reduceWhitespace
+        case (string, pattern) => Try { string.replaceAll(pattern, "").reduceWhitespace } match {
```
mdellabitta: getOrElse(string) instead of the pattern match?

src/main/scala/dpla/ingestion3/mappers/providers/OhioMapping.scala
```diff
@@ -114,13 +116,15 @@ class OhioMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeS
...
     extractStrings(data \ "metadata" \\ "dataProvider")
       .map(nameOnlyAgent)
-      .headOption // take the first value
-      .getOrElse( // return the first value or throw an exception
-      throw new Exception(s"Missing required property metadata/dataProvider is empty for ${getProviderId(data)}")
-    )
+      .headOption match {
+        case Some(dp) => dp
+        case None => msgCollector.add(missingRequiredError(getProviderId(data), "dataProvider"))
+          nameOnlyAgent("")  // FIXME this shouldn't have to return an empty value.
```
mdellabitta: Dumb question: Should this just throw if we can't assign a data provider? Or I guess, in the new world, log a fatal message?

src/main/scala/dpla/ingestion3/mappers/providers/OhioMapping.scala
```diff
@@ -129,20 +133,35 @@ class OhioMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeS
...
+      Try { new URI(uriStr)} match {
+        case Success(uri) => Option(uriOnlyWebResource(uri))
+        case Failure(f) =>
+          msgCollector.add(
+            mintUriError(id = getProviderId(data), field = "isShownAt", value = uriStr))
+          None
+      }
+    }).headOption match {
+      case None => msgCollector.add(missingRequiredError(getProviderId(data),"isShownAt")) // record error message
+        uriOnlyWebResource(new URI("")) // FIXME model should be changed to not compel empty EdmWebResource
```
mdellabitta: I think this would be fatal too.

src/main/scala/dpla/ingestion3/mappers/providers/WiMapping.scala
```diff
@@ -98,39 +100,47 @@ class WiMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq
...
+        case Success(uri) => Option(uriOnlyWebResource(uri))
+        case Failure(f) =>
+          msgCollector.add(mintUriError(id = getProviderId(data),field = "isShownAt", value = uriStr))
+          None
+      }
+    }).headOption match {
+      case Some(s) => s
+      case None =>
+        msgCollector.add(missingRequiredError(getProviderId(data), "isShownAt"))
+        uriOnlyWebResource(new URI("")) // TODO Fix this -- it requires an Exception thrown or empty EdmWebResource
```
mdellabitta: This would be Fatal too

src/main/scala/dpla/ingestion3/mappers/providers/WiMapping.scala
```diff
@@ -98,39 +100,47 @@ class WiMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq
...
+  override def preview(data: Document[NodeSeq])
+                      (implicit msgCollector: MessageCollector[IngestMessage]): ZeroToOne[EdmWebResource] =
+    extractString(data \ "metadata" \\ "preview").map(uriStr =>
+      Try(new URI(uriStr)) match {
         case Success(u) => uriOnlyWebResource(u)
-        case Failure(f) => throw new RuntimeException(s"Error mapping preview in record ${getProviderId(data)}. " +
-          s"Unable to mint URI from '$x': ${f.getMessage}")
+        case Failure(_) =>
+          msgCollector.add(mintUriError(id = getProviderId(data), field = "preview", value = uriStr))
+          uriOnlyWebResource(new URI(""))
```
mdellabitta: Should this just be None?

src/main/scala/dpla/ingestion3/model/DplaMapData.scala
mdellabitta: Maybe something we can turn off...

src/main/scala/dpla/ingestion3/mappers/providers/OhioMapping.scala
moltude: Yeah, that is what 125 is doing. `errors` are treated as fatal and I chose not to raise an exception because that would end the collection of messages for a record and I was striving to collect all messages (if a mapper has implemented them).

src/main/scala/dpla/ingestion3/mappers/providers/OhioMapping.scala
moltude: Yeah I need to change the level of these errors and bump it to fatal

src/main/scala/dpla/ingestion3/mappers/providers/WiMapping.scala
moltude: Good call

src/main/scala/dpla/ingestion3/mappers/providers/OhioMapping.scala
mdellabitta: Yeah I think I understand the problem a lot better now that we talked.

src/main/resources/avro/MAPRecord.avsc
moltude: Should have something on this by the end of the day

moltude: This is addressed in #181

PR #179, opened 2018-06-19
WI QA fixes (IN-453  and IN-463)

moltude:
* Add `limitCharacters(1000)` to description mapping
* Add prefix filtering to edmRights mapping to only map `edm:rights`
* Add `extent` field to extent mapping
* Fixup so that more than that the first collection title is mapped

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/WiMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/WiMapping.scala
@@ -33,7 +33,7 @@ class WiMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq
   override def alternateTitle(data: Document[NodeSeq]): Seq[String] = extractStrings(data \ "metadata" \\ "alternative")
 
   override def collection(data: Document[NodeSeq]): Seq[DcmiTypeCollection] =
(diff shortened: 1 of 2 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/WiMapping.scala
```diff
@@ -105,9 +107,17 @@ class WiMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq
...
     // Will quietly drop any invalid URIs
-    extractStrings(data \ "metadata" \\ "rights").find(r => Utils.isUri(r)).map(new URI(_))
+    val edmRights = (data \ "metadata" \\ "rights").map(rights => {
+      rights.prefix match {
+        case "edm" => rights.text
+        case _ => ""
+      }
+    })
+
+    edmRights.find(Utils.isUri).map(new URI(_))
```
AudreyAltman: Sanity check - will the statement on line 119 drop empty strings?  I assume that empty strings count as invalid URIs so they will be dropped, correct?

moltude: `find` returns the first value that meets the condition, `Utils.isUri` but I'm now noticing that the isUri method does not check for empty strings so this would result in an empty URI if the first value was an empty string. I'll fixup the `isUri` and this should be good.

Nice catch

PR #180, opened 2018-06-22
Add support for ISO-639-1 in Language enrichment

moltude:
* Add language resource file for iso-639-1 language codes
* Add test for iso-639 value
* Add ISO-639-1 to the default language enrichment lookup values
* Remove `enc,En` language code from iso639-3 resource file. Was able to consistently replicate behavior where a providedLabel code `en` would match iso639 code`enc`. I was also able to replicate the same bad matching on `env`, `enp`, and `enx` but not `enz`.  From my data review it appears that this was the only code creating bad data for Wisconsin.

### DIFF
--- a/src/main/resources/languages/iso639-1.csv
+++ b/src/main/resources/languages/iso639-1.csv
@@ -1,0 +1,206 @@
+# See http://www.loc.gov/standards/iso639-2/
+# code, English label
+aa,Afar
+ab,Abkhazian
+ae,Avestan
+af,Afrikaans
+ak,Akan
+am,Amharic
+an,Aragonese
+ar,Arabic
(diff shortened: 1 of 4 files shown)


moltude:
Finally! None of my local test runs failed but this Travis build was finally able to replicate the behavior we see in some of the records in CQA 
http://cqai3-1:8080/qa/compare?id=861000692a8db0af7d3b8fa0c85746b0

At least this proves it isn't a fluke and I suspect has something to do with the order of the vocab terms not being guarenteed so something like that.

moltude:
Simply rerunning the Travis builds is resolving this error so there is something clear wrong here and replicating has been tricky/

PR #181, opened 2018-06-25
Fix/lust4log enum

moltude:
Addresses issues raised by @mdellabitta in #176 about using enum in lieu of Strings for the logging level. 

I'm not 100% sure this was the correct implementation of that idea but its working and does the value checking of the field property.

### DIFF
--- a/src/main/scala/dpla/ingestion3/messages/IngestMessage.scala
+++ b/src/main/scala/dpla/ingestion3/messages/IngestMessage.scala
@@ -1,5 +1,13 @@
 package dpla.ingestion3.messages
 
+
+object IngestLogLevel {
+  val info: String = "info"
+  val warn: String = "warn"
+  val error: String = "error"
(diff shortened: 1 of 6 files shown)


src/main/scala/dpla/ingestion3/messages/IngestMessage.scala
```diff
@@ -1,5 +1,13 @@
 package dpla.ingestion3.messages
 
+
+object IngestLogLevel {
+  val info: String = "info"
+  val warn: String = "warn"
+  val error: String = "error"
+  val fatal: String = "fatal"
+}
```
AudreyAltman: Is the purpose of making these `String` types (as opposed to `Boolean` for example) that it makes logging output easier?

moltude: Partially, but tbh I just never considered anything than String

PR #182, opened 2018-06-25
Feature/log map enrich reports

moltude:
Adds proof of concept reporting to enrichment chain. Only providing "detail" reporting for the language field.


src/main/scala/dpla/ingestion3/executors/EnrichExecutor.scala
```diff
@@ -92,33 +102,65 @@ trait EnrichExecutor extends Serializable {
...
+    logger.info(EnrichmentSummary.getSummary(summaryData))
+
     sc.stop()
 
     // Clean up checkpoint directory, created above
     Utils.deleteRecursively(new File("/tmp/checkpoint"))
 
     // Log error messages.
-    failures.foreach(msg => logger.warn(s"Enrichment error >> $msg"))
+    failures.foreach(logger.error(_))
```
AudreyAltman: What does line 140 do?  Does it write to a log file?

src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
```diff
@@ -76,91 +79,120 @@ trait MappingExecutor extends Serializable {
...
-
-    logger.info(MappingSummary.getSummary(mappingSummary))
-    logger.info(s"Number of exceptions ${failures.length}")
-
-    // Write warn and error messages to CSV files. These should all share the same timestamp, minor work TBD
     val baseLogDir = s"$dataOut/../logs/"
-    val time = timeInMs.toString
 
-    val exceptions = sc.parallelize(failures).toDS()
+    val exceptionsDS = sc.parallelize(exceptions).toDS()
```
AudreyAltman: Do you need to make this into a `DataSet`?  Would it be easier to work with it as a plain `RDD`?  Not a big deal, just curious.

src/main/scala/dpla/ingestion3/reports/summary/ReportFormattingUtils.scala
```diff
@@ -0,0 +1,14 @@
+package dpla.ingestion3.reports.summary
+
+import org.apache.commons.lang.StringUtils
+
+/**
+  *
+  */
```
AudreyAltman: Empty comment block.

src/main/scala/dpla/ingestion3/executors/EnrichExecutor.scala
moltude: Exactly

PR #183, opened 2018-06-26
Add type field to enrichment field reports

moltude:
- Refactor generateLanguageReport into a  generic method for all fields
- Fixup formatting of messages in `MAPRecord.avsc`
- Matching on the type field is done by zipping the two arrays together and comparing the values of each tuple. For example:
  TypeA = (a,b,a)
  TypeB = (A,b,A)
  zipAB = ((a,A), (b,b), (a,A))

 Values are compared and produce the following:
  (a,A) = enriched
  (b,b) = not enriched
  (a,A) = enriched


src/main/scala/dpla/ingestion3/reports/PrepareEnrichmentReport.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  def generateFieldReport(ds: Dataset[Row], field: String) = {
+    val queryResult = ds
+      .select("level", "message", "id", "field")
+      .where(s"field=='$field'")
+      .drop("field")
+      .groupBy("level", "message")
+      .agg(count("id")).as("count")
+      .orderBy("level","message")
+      .orderBy(desc("count(id)"))
```
AudreyAltman: Could this be one `orderBy` statement instead of two?  Not a big deal, mostly I'm just curious.

src/main/scala/dpla/ingestion3/messages/IngestMessageTemplates.scala
```diff
@@ -0,0 +1,48 @@
+package dpla.ingestion3.messages
+
+import java.net.URI
+
+import scala.util.{Failure, Success, Try}
+
+trait IngestMessageTemplates {
```
AudreyAltman: I like that you templated this.  These are good helper methods.

src/main/scala/dpla/ingestion3/mappers/providers/PaMapping.scala
```diff
@@ -78,12 +81,12 @@ class PaMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq
...
-    if (contributors.nonEmpty)
-      nameOnlyAgent(contributors.last)
-    else
-      throw new Exception(s"Missing required property dataProvider because " +
-        s"dc:contributor is empty for ${getProviderId(data)}")
+    extractStrings(data \ "metadata" \\ "contributor").lastOption match {
+      case Some(lastContributor) => nameOnlyAgent(lastContributor)
+      case None =>
+        msgCollector.add(missingRequiredError(getProviderId(data), "dataProvider"))
+        nameOnlyAgent("") // FIXME this shouldn't have to return an empty value.
```
AudreyAltman: Just flagging this `FIXME` comment in case you meant to fix it sooner rather than later.

src/main/scala/dpla/ingestion3/reports/PrepareEnrichmentReport.scala
moltude: I thought it complained about something when I put it on a single line but I might be misremembering.

PR #190, opened 2018-07-24
Add South Dakota 

moltude:
Add South Dakota to ingestion3 using mapper based on existing MDL mapping and MDL harvester

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/SdMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/SdMapping.scala
@@ -1,0 +1,22 @@
+package dpla.ingestion3.mappers.providers
+
+import java.net.URI
+
+import dpla.ingestion3.mappers.utils._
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/SdMapping.scala
```diff
@@ -0,0 +1,22 @@
...
+
+// FIXME Why is the implicit conversion not working for JValue when it is for NodeSeq?
+class SdMapping extends MdlMapping with JsonMapping with JsonExtractor with IdMinter[JValue] {
+
+  // ID minting functions
+  override def useProviderName: Boolean = true
+
+  override def getProviderName: String = "sd"
+
+  override def agent = EdmAgent(
```
moltude: Just noticed that `agent` is redundant and could be removed since it isn't actually overriding anything

moltude:
@AudreyAltman ready

PR #191, opened 2018-07-27
Addresses IN-476 - Begin/End dates

moltude:
Add support for including begin and end dates at the level of precision of the original date. For example:
    * May 1959 --> 1949-05
    * 1922 --> 1922
    * Oct 5 1821 --> 1821-10-05
    * 11-1952 --> 1952-11

### DIFF
--- a/src/main/scala/dpla/ingestion3/enrichments/date/DateBuilder.scala
+++ b/src/main/scala/dpla/ingestion3/enrichments/date/DateBuilder.scala
@@ -1,37 +1,34 @@
 package dpla.ingestion3.enrichments.date
 
-import org.joda.time.LocalDate
-import org.joda.time.format.{DateTimeFormat, DateTimeFormatter, DateTimeFormatterBuilder}
-
+import java.text.SimpleDateFormat
+import dpla.ingestion3.model._
 import scala.util.{Failure, Success, Try}
 
 /**
   *
-  * @param regexPattern Patten to match against
-  * @param dateTimePattern Pattern to transform into
-  * @param label Label for this pattern to pattern transformation
+  * @param parser
+  * @param format
   */
-case class EdtfPatternMap(regexPattern: String, dateTimePattern: String, label: String)
+case class EdtfPatternMap(parser: SimpleDateFormat, format: SimpleDateFormat)
 
 /**
   *
   */
 object DateBuilderPatterns {
-  // Support date patterns
-  val dateTimeFormatPatterns = Array(
-    // Single date
-    DateTimeFormat.forPattern("yyyy").getParser,
-    DateTimeFormat.forPattern("yyyy MM").getParser,
-    DateTimeFormat.forPattern("yyyy MM dd").getParser,
-    DateTimeFormat.forPattern("yyyy MMM").getParser,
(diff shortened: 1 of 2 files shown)


src/main/scala/dpla/ingestion3/enrichments/date/DateBuilder.scala
```diff
@@ -1,37 +1,34 @@
...
-    DateTimeFormat.forPattern("yyyy MMM dd").getParser,
-    DateTimeFormat.forPattern("MMM yyyy").getParser
 
+  val edtfPatterns = Seq(
+    EdtfPatternMap(new SimpleDateFormat("yyyy MM dd"),  new SimpleDateFormat("yyyy-MM-dd")),
+    EdtfPatternMap(new SimpleDateFormat("yyyy MMM dd"),    new SimpleDateFormat("yyyy-MM-dd")),
+    EdtfPatternMap(new SimpleDateFormat("MMM yyyy"),    new SimpleDateFormat("yyyy-MM")),
+    EdtfPatternMap(new SimpleDateFormat("yyyy MMM"),    new SimpleDateFormat("yyyy-MM")),
+    EdtfPatternMap(new SimpleDateFormat("yyyy MM"),     new SimpleDateFormat("yyyy-MM")),
+    EdtfPatternMap(new SimpleDateFormat("yyyy"),        new SimpleDateFormat("yyyy"))
```
AudreyAltman: I'm wondering if there is a way to generalize this.  Here you are explicitly stating each `parser` pattern and associating it with a `format` pattern.  But the `format` pattern is pretty much the same every time, except that it knows  if year, month, and date are present.  Would there be a way to abstract this so that you could universally apply the `format` pattern to any `parser` without the `format` having to know if year, month, and date are present in the `parser`?  Then there could just be one `parser` that works on any `format`.

moltude: I'm not sure if that is possible at the moment but I might not be thinking through all the options. I can identify two possible constraints that I've run into that might prevent implementing something like that. 

1. Using a universal format pattern like `yyyy-MM-dd` for all parsers when the parser is set to `yyyy-MM` would automatically pad out the date to `yyyy-MM-dd` (2015-11 becomes 2015-11-01) and that is something we were actively trying to avoid here. We wanted the granularity of the begin and end dates for a single date to align with the original date. So rather than strip those extraneous values off after the date has been formatted I thought it was better to match granularity when pairing `pattern` to `format` 
2. The other issue (though this may not directly apply to your suggestion) is that the order of the `patterns` is important so that it will match `yyyy MM` before `yyyy`.  So far as I've been able to determine these parsers cannot be told "match on full pattern" and so if `yyyy` was the first pattern then nearly all dates would match. They have to be ordered from most specific to least specific to achieve the desired result. 

What do you think? Am I missing something here? I would like to not have to explicitly map out all the patterns and formats but in the moment I didn't see much of a way around these problems.

AudreyAltman: That all makes sense.  It was just an idea, but it sounds like you have good reasons for implementing it the way you did.

moltude:
Thanks Audrey

PR #193, opened 2018-08-09
Feature/in 417 folder deletion

mdellabitta:
- Moved folder deletion call to entry.

- Refactored so that harvests that write to avro directly
extend an abstract class that handles the AvroWriter
lifecycle, and everything else extends Harvester.

### DIFF
--- a/src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
+++ b/src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
@@ -1,5 +1,7 @@
 package dpla.ingestion3.entries.ingest
 
+import java.io.File
+
 import dpla.ingestion3.confs.{CmdArgs, Ingestion3Conf}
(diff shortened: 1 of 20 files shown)


src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
```diff
@@ -51,6 +53,13 @@ object IngestRemap extends MappingExecutor
       .set("spark.kryoserializer.buffer.max", "200")
       .setMaster(sparkMaster)
 
+
+    Utils.deleteRecursively(new File(harvestDataOut))
```
moltude: @mdellabitta I don't think we want to do this here ;)

mdellabitta: :man_facepalming:

mdellabitta: I was just being thorough!

PR #198, opened 2018-08-21
MT Mapping

moltude:
Add MT mapping

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
@@ -1,0 +1,148 @@
+package dpla.ingestion3.mappers.providers
+
+import java.net.URI
+
+import dpla.ingestion3.enrichments.normalizations.StringNormalizationUtils._
+import dpla.ingestion3.enrichments.normalizations.filters.{DigitalSurrogateBlockList, FormatTypeValuesBlockList}
+import dpla.ingestion3.mappers.utils.{Document, IdMinter, Mapping, XmlExtractor}
+import dpla.ingestion3.messages.{IngestMessage, MessageCollector}
+import dpla.ingestion3.model.DplaMapData.{ExactlyOne, ZeroToOne}
+import dpla.ingestion3.model._
+import dpla.ingestion3.utils.Utils
+import org.json4s.JValue
+import org.json4s.JsonDSL._
+
+import scala.xml._
+
+
+class MtMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq] {
+
+  val formatBlockList: Set[String] =
+    DigitalSurrogateBlockList.termList ++
+      FormatTypeValuesBlockList.termList
+
+  // ID minting functions
+  override def useProviderName(): Boolean = true
+
+  override def getProviderName(): String = "mt"
+
+  override def getProviderId(implicit data: Document[NodeSeq]): String =
+    extractString(data \ "header" \ "identifier")
+      .getOrElse(throw new RuntimeException(s"No ID for record $data")
+      )
+
+  // SourceResource mapping
+  override def collection(data: Document[NodeSeq]): Seq[DcmiTypeCollection] =
+    // first instance of <mods:relatedItem><mods:titleInfo><mods:title>
+    extractStrings(data \\ "relatedItem" \ "titleInfo" \ "title")
+      .map(nameOnlyCollection)
+      .take(1)
+
+  override def creator(data: Document[NodeSeq]): Seq[EdmAgent] =
+    // <mods:name><mods:namePart> when <mods:role><mods:roleTerm> equals Creator
+  (data \\ "name")
+    .map(node => node.filter(n => (n \\ "roleTerm").text.equalsIgnoreCase("creator")))
+    .flatMap(n => extractStrings(n \ "namePart"))
+    .map(nameOnlyAgent)
+
+  override def date(data: Document[NodeSeq]): Seq[EdmTimeSpan] =
+  // <mods:originInfo><mods:dateCreated>
+    extractStrings(data \\ "originInfo" \ "dateCreated")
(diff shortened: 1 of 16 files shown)


src/main/scala/dpla/ingestion3/profiles/ProviderProfiles.scala
```diff
@@ -57,6 +57,17 @@ class MdlProfile extends JsonProfile {
   override def getMapping = new MdlMapping
 }
 
+/**
+  * (Montana) Big Sky Digital Network
+  */
+class MtProfile extends XmlProfile {
+  type Mapping = OhioMapping
```
AudreyAltman: Why is the Mapping type `OhioMapping`?

src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
```diff
@@ -0,0 +1,169 @@
...
+    DigitalSurrogateBlockList.termList ++
+      ExtentIdentificationList.termList
+
+  // ID minting functions
+  override def useProviderName(): Boolean = true
+
+  override def getProviderName(): String = "mt"
+
+  override def getProviderId(implicit data: Document[NodeSeq]): String =
+    extractString(data \\ "header" \ "identifier").getOrElse(throw new RuntimeException(s"No ID for record $data"))
```
AudreyAltman: Why do you throw a runtime exception here instead of adding an error to the `msgCollector`?

src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
```diff
@@ -0,0 +1,169 @@
...
+    extractStrings(data \\ "originInfo" \ "dateCreated")
+      .map(stringOnlyTimeSpan)
+
+  override def description(data: Document[NodeSeq]): Seq[String] =
+  // <mods:note> @type=content
+    (data \\ "metadata" \ "mods" \ "note")
+      .flatMap(n => getByAttribute(n.asInstanceOf[Elem], "type", "content"))
+      .flatMap(n => extractStrings(n))
+
+  override def extent(data: Document[NodeSeq]): ZeroToMany[String] =
```
AudreyAltman: I notice that you specify a return type of `ZeroToMany[String]` here but in other places, such as on line 61, you specify a return type of `Seq[String]`.  Is there a rule of thumb for when to use the custom types?

src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
```diff
@@ -0,0 +1,169 @@
...
+  override def dplaUri(data: Document[NodeSeq]): URI = mintDplaItemUri(data)
+
+  override def dataProvider(data: Document[NodeSeq])
+                           (implicit msgCollector: MessageCollector[IngestMessage]): EdmAgent =
+  // <mods:note> @type=ownership
+  (data \\ "metadata" \ "mods" \ "note")
+    .flatMap(node => getByAttribute(node.asInstanceOf[Elem], "type", "ownership"))
+    .flatMap(extractStrings)
+    .map(nameOnlyAgent)
+    .headOption match {
```
AudreyAltman: For readability, you could indent the lines after line 106.

src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
```diff
@@ -0,0 +1,169 @@
...
+      nameOnlyAgent("") // FIXME this shouldn't have to return an empty value.
+  }
+
+  override def edmRights(data: Document[NodeSeq]): ZeroToOne[URI] = {
+    // <mods:accessCondition>@type=use and reproduction @xlink:href =[this is the value to be mapped]
+    (data \\ "metadata" \ "mods" \ "accessCondition")
+      .flatMap(node => getByAttribute(node.asInstanceOf[Elem], "type", "use and reproduction"))
+      .flatMap(node => node.attribute(node.getNamespace("xlink"), "href"))
+      .flatMap(n => extractString(n.head))
+      .headOption match {
```
AudreyAltman: Another place where you could indent, after line 119.

src/test/scala/dpla/ingestion3/mappers/providers/MtMappingTest.scala
```diff
@@ -0,0 +1,122 @@
...
+//    val xml: Document[NodeSeq] = Document(
+//    <record xmlns:dc="http://purl.org/dc/elements/1.1/">
+//      <metadata>
+//        <dc:rights>Eric Rights Holder</dc:rights>
+//        <edm:rights>http://rights.holder.edu</edm:rights>
+//      </metadata>
+//    </record>)
+//    val expected = Option(new URI("http://rights.holder.edu"))
+//    assert(extractor.edmRights(xml) === expected)
+//  }
```
AudreyAltman: Did you mean to keep these commented-out sections of the test?

src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
moltude: That was just the IntelliJ auto complete function filling in `Seq` instead of `ZeroToMany`

moltude:
@AudreyAltman Made your requested changes

src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
moltude: The `IdMinter` trait doe not currently extend `IngestMessageTemplates` which it would need to do for the `getProviderId` and other id-related methods in the mappings to log message templates. This gap should probably also be fixed in a different way rather than depending on every mapping to log the message. The better way would probably have the `Mapper` or `Xml/JsonMapper` check for required properties rather than reimplement the same messaging code for every single mapping.

AudreyAltman: :+1:

PR #199, opened 2018-08-28
Initial IA harvester

moltude:
Internet Archive harvester. Pulls from the Search API

### DIFF
--- a/src/main/scala/dpla/ingestion3/harvesters/api/IaHarvester.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/api/IaHarvester.scala
@@ -2,11 +2,12 @@ package dpla.ingestion3.harvesters.api
 
 import java.net.URL
 
+import com.databricks.spark.avro._
 import dpla.ingestion3.confs.i3Conf
 import dpla.ingestion3.utils.HttpUtils
 import org.apache.http.client.utils.URIBuilder
 import org.apache.log4j.Logger
-import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.json4s.DefaultFormats
 import org.json4s.JsonAST.JValue
 import org.json4s.jackson.JsonMethods.{compact, parse, render}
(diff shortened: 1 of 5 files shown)


src/main/scala/dpla/ingestion3/harvesters/api/IaHarvester.scala
```diff
@@ -0,0 +1,121 @@
...
+            error.errorSource.url.getOrElse("Undefined url"),
+            error.errorSource.queryParams,
+            error.message
+          ))
+          continueHarvest = false
+        // Handle a successful response
+        case src: ApiSource with ApiResponse =>
+          src.text match {
+            case Some(docs) =>
+              val json = parse(docs)
```
AudreyAltman: Do you want to handle the possibility here that `parse(docs)` will throw an error if IA's JSON is malformed?  Or will it just bubble up in some useful way?  May not be worth worrying about if it's a low risk.  For that matter, is there a chance of an exception being thrown if the `identifier` and `cursor` aren't present where you expect them in the next few lines?

src/main/scala/dpla/ingestion3/profiles/ProviderProfiles.scala
```diff
@@ -36,6 +36,15 @@ class EsdnProfile extends XmlProfile {
   override def getMapping = new EsdnMapping
 }
 
+/**
+  * Internet Archive
+  */
+class IaProfile extends JsonProfile {
+  type Mapping = CdlMapping // FIXME
```
AudreyAltman: Do you want to keep `CdlMapping` on lines 43 and 46 for now?

src/main/scala/dpla/ingestion3/harvesters/api/IaHarvester.scala
```diff
@@ -0,0 +1,121 @@
...
+
+              // Loop control
+              cursor = (json \\ "cursor").extractOrElse[String]("")
+
+              if (cursor.isEmpty)
+                continueHarvest = false
+            // Handle unknown case
+            case _ =>
+              harvestLogger.error(s"Response body is empty.\n" +
+                s"URL: ${src.url.getOrElse("!!! URL not set !!!")}\n" +
```
AudreyAltman: I like the exclamation marks :)

src/main/scala/dpla/ingestion3/profiles/ProviderProfiles.scala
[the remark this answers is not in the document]
moltude: Yes, this is just a place holder until the Mapping is written. `type Mapping = ???` isn't valid and using a stub like this is the only work around that I've found so far.

moltude: Yes, see above

src/main/scala/dpla/ingestion3/harvesters/api/IaHarvester.scala
moltude: I added some additional error handling in the latest commit

moltude:
@AudreyAltman bumping for changes

src/main/scala/dpla/ingestion3/profiles/ProviderProfiles.scala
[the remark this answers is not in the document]
mdellabitta: Maybe we just need an EvilMapping that throws for every method called on it as a placeholder.

PR #201, opened 2018-09-04
Automate harvest output path

AudreyAltman:
This PR should not be reviewed/merged until https://github.com/dpla/ingestion3/pull/200 has been approved.

This creates a util method to generate output paths for ingest activities.  It is only implemented for harvests in this PR, but can be expanded to encompass all other activities as well.

With this workflow, the user specifies the top-level directory or S3 bucket, and the util method generates sub-directories.  For example, if a user did:

```
sbt "run-main dpla.ingestion3.entries.ingest.HarvestEntry 
  --conf=my-conf 
  --output=s3://my-bucket/  
  --name=esdn"
```

then the util method would generate the following output path for a harvest (with the correct timestamp):

`s3://my-bucket/esdn/harvest/20180904_150911-esdn-OriginalRecord.avro`

To run a production harvest, the user would simply pass in `--output=s3://dpla-master-dataset/ `.  To run a local harvest, the user would pass in `--output=/local/filepath`.


AudreyAltman:
@moltude I agree that deleting old files in the file path is probably not necessary.  I deleted it, and if it becomes a problem we can add in a flag as you suggested.

PR #205, opened 2018-09-10
Manifest writer

AudreyAltman:
This adds a manifest writing capabilities to the `OutputHelper` and implements them for harvests only.  It also adds some unit tests.

Out of scope for this PR is moving FileIO functionality to more general utils, or making use of what utils already exist.  This will be the next PR.

I tested this writing locally and to s3.  Here is an example manifest file:

```
Activity: Harvest
Provider: esdn
Record count: 2163
Start date/time: 2018-09-10 14:34:18
```

### DIFF
--- a/src/main/scala/dpla/ingestion3/utils/OutputHelper.scala
+++ b/src/main/scala/dpla/ingestion3/utils/OutputHelper.scala
@@ -1,38 +1,173 @@
 package dpla.ingestion3.utils
 
-import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
+import java.time.LocalDateTime
+import java.io.{BufferedWriter, File, FileWriter}
 
-object OutputHelper {
+import com.amazonaws.services.s3.model.{PutObjectRequest, PutObjectResult}
+import com.amazonaws.services.s3.AmazonS3Client
+
+import scala.util.Try
+
+/*
+ * @param root: Root directory or AWS S3 bucket.
+ *              If `root' is an AWS S3 bucket, it should start with "s3a://".
+ * @param shortName: Provider short name
+ * @param activity: "harvest", "mapping", "enrichment", etc.
+ * @param startDateTime: start dateTime of the activity
+ *
+ * @throws IllegalArgumentException
+ *
+ * @see https://digitalpubliclibraryofamerica.atlassian.net/wiki/spaces/TECH/pages/84512319/Ingestion+3+Storage+Specification
+ *      for details on file naming conventions
+ *
+ */
+class OutputHelper(root: String,
+                   shortName: String,
+                   activity: String,
+                   startDateTime: LocalDateTime) {
 
   /*
-   * Get output path for a harvest, mapping, enrichment, or indexing activity.
-   *
-   * @example:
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/utils/OutputHelper.scala
```diff
@@ -1,38 +1,173 @@
...
-    val timestamp: String = LocalDateTime.now.format(formatter)
+
+    val timestamp: String = startDateTime.format(formatter)
 
     // TODO: handle "reports" case
     // TODO: handle "logs" case
-    val schema = activity match {
+    // TODO: make schema configurable?
+    val schema: String = activity match {
       case "harvest" => "OriginalRecord"
```
mdellabitta: This list of operations looks potentially like a good place to use a sealed case class hierarchy rather than strings. Then the case clasess could store things like the schema rather than having to spread this type of information around the project.

src/main/scala/dpla/ingestion3/utils/OutputHelper.scala
```diff
@@ -1,38 +1,173 @@
...
+    file.getParentFile.mkdirs
+    val bw: BufferedWriter = new BufferedWriter(new FileWriter(file))
+    bw.write(text)
+    bw.close()
+    file
+  }
+
+  // Write a String to an S3 file.
+  // TODO: Move to a more general FileIO Util
+  def writeS3File(bucket: String, key: String, text: String): PutObjectResult = {
```
mdellabitta: I'd probably not move this method somewhere else, and I'd also find another return type than PutObjectResult. Think of this method as a facade over writing to S3, which should make it so callers don't have to think about S3 at all.

src/main/scala/dpla/ingestion3/utils/OutputHelper.scala
```diff
@@ -1,38 +1,173 @@
...
+    // Write to a temporary local file
+    val tempOut: String = s"$tempDirectory/$key"
+    val tempFile: File = writeLocalFile(tempOut, text)
```
mdellabitta: `putObject` has an alternative implementation that takes an InputStream, so you don't need to save locally first. It would go something like:
```
val in = new ByteArrayInputStream(text.getBytes("utf-8"))
s3client.putObject(new PutObjectRequest(bucket, key, in, new ObjectMetadata())
```

AudreyAltman: @mdellabitta I saw that but I was did not understand what `ObjectMetadata` was all about.  Have you used it before?

mdellabitta: It's sort of like HTTP headers... you can set a mime type, etc. Not especially relevant.

AudreyAltman:
@mdellabitta I believe I addressed all your comments:
* Using input stream instead of temp files when writing to S3
* Returning a `Try[String]` for both local and S3 writes (instead of `PutObjectResult` for S3 write and `File` for local write)
* Added comment about using closed case class for activities so it can be implemented in later PR

PR #209, opened 2018-09-11
IngestRemap to use most recent harvested data for mapping and enrichment

moltude:
Add method getMostRecent(path: String) to Utils which will return the absolute path to the most recent data folder
IngestRemap uses getMostRecent() to select data to map and enrich

### DIFF
--- a/src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
+++ b/src/main/scala/dpla/ingestion3/entries/ingest/IngestRemap.scala
@@ -30,15 +30,19 @@ object IngestRemap extends MappingExecutor
     val confFile = cmdArgs.getConfigFile()
     val shortName = cmdArgs.getProviderName()
 
+    // Get logger
+    val logger = Utils.createLogger("ingest", shortName)
+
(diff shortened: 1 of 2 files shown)


mdellabitta:
Is this redundant with stuff Audrey's working on?

AudreyAltman:
I'd like to implement something like this, but it might be best to wait until a couple other things have been finalized.  Okay to let this PR sit for just a little while and then come back to it once other stuff has been solidified?

moltude:
Yeah, I just need it for running re-ingests, having to lookup and retype the harvest output dir became tedious.

mdellabitta:
@AudreyAltman Since it's helpful now, why don't we merge it and then when you revisit later, we can yank it? It's not a big change?

AudreyAltman:
Oh sure, if it's useful now then go for it.

AudreyAltman:
@moltude I think this is fine, I would just caution you to be mindful that the output directory structure may change with the stuff I've been working on - so make sure the harvests are where you think they are so you can pass in the correct root directory.

PR #212, opened 2018-09-14
Updates to MDL mapping

moltude:
Add edmRights mapping and test
Add blockFilter to format mapping 
Add intermediateProvider mapping 
Fixup language, place, publisher, subject mappings

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/MdlMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/MdlMapping.scala
@@ -2,13 +2,14 @@ package dpla.ingestion3.mappers.providers
 
 import java.net.URI
 
+import dpla.ingestion3.enrichments.normalizations.StringNormalizationUtils._
+import dpla.ingestion3.enrichments.normalizations.filters.ExtentIdentificationList
 import dpla.ingestion3.mappers.utils._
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/MdlMapping.scala
```diff
@@ -39,6 +42,15 @@ class MdlMapping extends JsonMapping with JsonExtractor with IdMinter[JValue] wi
   override def dplaUri(data: Document[JValue]): ExactlyOne[URI] =
     new URI(mintDplaId(data))
 
+  override def edmRights(data: Document[json4s.JValue]): ZeroToOne[URI] =
+    extractString(unwrap(data)  \\ "record" \ "rights") match {
+      case Some(t) => Some(new URI(t))
+      case None => None
+    }
+
+  override def intermediateProvider(data: Document[_root_.org.json4s.JsonAST.JValue]): ZeroToOne[EdmAgent] =
```
AudreyAltman: Question about the data input value: I notice that here the value `data` is defined as `Document[_root_.org.json4s.JsonAST.JValue]`, while in the method above it is `Document[json4s.JValue]` and below it is `Document[JValue]`.  What is the difference?

PR #213, opened 2018-09-25
MT mapping fixes

moltude:
Add splitAtDelimiter to format, place, subject mappings

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
@@ -3,7 +3,7 @@ package dpla.ingestion3.mappers.providers
 import java.net.URI
 
 import dpla.ingestion3.enrichments.normalizations.StringNormalizationUtils._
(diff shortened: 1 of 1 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/MtMapping.scala
```diff
@@ -87,6 +89,7 @@ class MtMapping extends Mapping[NodeSeq] with XmlExtractor with IdMinter[NodeSeq
   override def subject(data: Document[NodeSeq]): ZeroToMany[SkosConcept] =
   // <mods:subject><mods:topic>
     extractStrings(data \\ "subject" \ "topic")
+      .flatMap(_.splitAtDelimiter(";"))
```
AudreyAltman: @moltude what is the difference between `split` and `splitAtDelimiter`?

moltude: applies trim and filter 
```
    lazy val splitAtDelimiter: (String) => Array[String] = (delimiter) => {
      value.split(delimiter).map(_.trim).filter(_.nonEmpty)
    }```

AudreyAltman: 👍

PR #218, opened 2018-10-05
Internet Archive mapping

moltude:
Addresses IN-557. Add initial Internet Archive mapping.


src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -131,5 +131,6 @@ case class EdmTimeSpan(
 
 case class URI(value: String) {
   def validate: Boolean = Try { new java.net.URI(value) }.isSuccess
-  def print: String = value
+
+  override def toString: String = value
```
AudreyAltman: Could you add a comment here explaining why you needed to do this?  I can imagine in six months I won't remember.

moltude: Done.

PR #221, opened 2018-10-05
Jsonl output

AudreyAltman:
This standardizes output for jsonl (i.e. indexing) jobs.

Some of the changes I made in this PR may warrant further discussion.

One issue I encountered was that the indexing job needs to be aware of the provider `shortName` in order to create the correct filepath.  Before, the indexer was blissfully ignorant of the provider `shortName`.  

This PR also changes `JsonlEntry` to accept named parameters, which brings it in line with all the other `Entry`s.  Please let me know if there is a reason I am unaware of that makes this change undesirable.  Also, in using the existing `CmdArgs`, I had to change the `conf` parameter from required to optional b/c `JsonlEntry` does not need a `conf`.  Again, this may not be desirable. 

This has been tested locally and on EC2.

### DIFF
--- a/src/main/scala/dpla/ingestion3/entries/ingest/JsonlEntry.scala
+++ b/src/main/scala/dpla/ingestion3/entries/ingest/JsonlEntry.scala
@@ -11,8 +11,8 @@ import org.apache.spark.SparkConf
   * Elasticsearch 0.90
   *
   * Expects three parameters:
-  * 1) a path to the harvested data
-  * 2) a path to output the mapped data
+  * 1) a path to the mapped/enriched data
+  * 2) a path to output the jsonl data
   * 3) provider short name (e.g. 'mdl', 'cdl', 'harvard')
   *
   * Usage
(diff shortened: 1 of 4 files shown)


src/main/scala/dpla/ingestion3/entries/ingest/JsonlEntry.scala
```diff
@@ -11,34 +10,35 @@ import org.apache.spark.SparkConf
...
-  * The output directory will contain one file named "part-*" that constitutes
-  * the JSONL.
-  *
-  *   Usage
-  *   -----
-  *   To invoke via sbt:
-  *     sbt "run-main dpla.ingestion3.entries.JsonlEntry /input/path /output/path"
+  * Expects three parameters:
+  * 1) a path to the harvested data
+  * 2) a path to output the mapped data
```
moltude: `a path to output the jsonl`

moltude: `a path to the enriched data`

PR #223, opened 2018-10-06
Fixup file names in _LOGS folder

moltude:
Addresses a nagging problem where the CSV files written out by spark have a generated file name 'part-xxxx-xxxx-xxxx-xxxx-xxxxx.csv'

For example:
`./ohio/mapping/20181005_230045-ohio-MAP4_0.MAPRecord.avro/_LOGS/ohio-123453453-map-errors/part-0000-0000-0000-1234-1234.csv`

Now when writing a DataFrame out as a CSV it is first written to a temporary file and then copied to its final location.

Now:
`./ohio/mapping/20181005_230045-ohio-MAP4_0.MAPRecord.avro/_LOGS/errors.csv`

There may be an argument for keeping that date/time information in the CSV filename so that we aren't losing any of the run information if the files in the `_LOGS` folder get copied out of data directory but I thought that this structure is a lot cleaner and easier to read since the data/time information between the log files and the data directory didn't line up (date_time vs time in milliseconds).


moltude:
*TODO This fix should also be used when writing out reports

mdellabitta:
I think the standard for Spark/Hadoop output is that the directory name is significant and the file name is not. Seems like this might end up being a slippery slope...

mdellabitta:
Meaning, we could do this for log files, but I'd start to get a little leery of doing it for actual data.

moltude:
I don't think I intended this for the data, but sending part-0000-0000-000-0000.csv to the providers is not what anyone would be expecting IMO.

mdellabitta:
I definitely think that if we start working against the abstractions built into this system we're using, we could very easily end up getting bitten. And I'm not sure that violating expectations is super harmful here. Just advocating for caution.

PR #239, opened 2018-10-26
Addresses IN-583

moltude:
- Add validation for recommended fields (logged as a warning)
- Add validation for required field title (logged as an error)


src/main/scala/dpla/ingestion3/mappers/Mapper.scala
```diff
@@ -135,6 +151,22 @@ trait Mapper[T, +E] extends IngestMessageTemplates {
...
+    * @param providerId String            The provider's local identifier
+    * @param collector MessageCollector   Ingest message collector
+    */
+  def validateTitle(titles: ZeroToMany[String], providerId: String)
+                   (implicit collector: MessageCollector[IngestMessage]): ZeroToMany[String] = {
+    if(titles.isEmpty) {
+      collector.add(missingRequiredError(providerId, "title"))
+    }
+    titles
+  }
```
AudreyAltman: Why validate `title` with this special `validateTitle` method rather than the generic `validateRecommendedProperty` method?  Could you document the reason in a comment?

mdellabitta: Title is required, I think.

moltude: Added comments to address comments

PR #240, opened 2018-10-29
Added default dataProvider per IN-587.

mdellabitta:


moltude:
@mdellabitta I see this as passing the tests and other than the quality complaint about `get` I think this is is all up to date AFAIK. I can't find any other open/unmerged IA branches on my local.

mdellabitta:
Hey @moltude I'm trying to figure out where you're seeing a warning about `.get`. Usually that's for `Option`, but is OK in this case, right? Right now Codacy is kinda being brain dead for me, and I can't see any warnings about this PR.

PR #241, opened 2018-11-06
Addresses IN-586 Updates to DPLA model

moltude:
- Add isReferencedBy field to EdmWebResource
- Update model and row converters with new field
- Update ModelConverterTest


moltude:
No problem to test, Ohio is enriching now but since its in mid-process the `row > model` pieces is working ;)

mdellabitta:
Just want to make sure we know ahead of time if we need to reprocess all the data prior to merging this.

moltude:
@mdellabitta Now works. Enriches older version of mapped data. Took more fiddling that I expected.

mdellabitta:
Nice, that looks super tidy!

PR #244, opened 2018-11-20
Digital virginias mapper

AudreyAltman:
This is the first draft of the Digital Virginias mapper.

Since there isn't a harvested dataset yet, this has only been validated through automated test against the sample `virginias.xml` file.

A few specific questions are detailed in `TODO` comments.

I am compiling a list of questions/communications for Digital Virginia, including:
* Is `dcterms:identifier` an appropriate field to use for a persistent identifier?
* There appears to be no values in the actual original records for `dc:temporal`, but it appears in the provider crosswalk.  Should it be included in the mapping?
* There appears to be no values in the actual original records for `dc:contributor`, but it appears in Gretchen's notes.  Should it be included in the mapping?
* We currently do not have the ability to map `edm:hasType` to `Subtype` even though it appears in the provider crosswalk.


src/main/scala/dpla/ingestion3/profiles/ProviderProfiles.scala
```diff
@@ -157,6 +157,17 @@ class SdProfile extends JsonProfile {
...
 }
 
+/**
+  * Digital Virginias
+  */
+class VirginiasProfile extends XmlProfile {
+  type Mapping = VirginiasMapping
+
+  // TODO Set correct harvester
+  override def getHarvester = classOf[OaiHarvester]
```
mdellabitta: Obviously need to revise this later...

PR #246, opened 2018-11-26
Va harvester

AudreyAltman:
This adds a file harvester for Digital Virginias.  It has been tested locally.

### DIFF
--- a/src/main/scala/dpla/ingestion3/harvesters/file/NaraFileHarvester.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/file/NaraFileHarvester.scala
@@ -123,9 +123,8 @@ class NaraFileHarvester(
         FileResult(entry.getName, result) #:: iter(tarInputStream)
     }
 
-
   /**
-    * Executes the plains2peaks harvest
+    * Executes the nara harvest
     */
   override def localHarvest(): DataFrame = {
     val harvestTime = System.currentTimeMillis()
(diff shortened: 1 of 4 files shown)


src/main/scala/dpla/ingestion3/harvesters/file/NaraFileHarvester.scala
```diff
@@ -123,9 +123,8 @@ class NaraFileHarvester(
         FileResult(entry.getName, result) #:: iter(tarInputStream)
     }
 
-
   /**
-    * Executes the plains2peaks harvest
+    * Executes the nara harvest
```
mdellabitta: Uh nice snag! :)

src/main/scala/dpla/ingestion3/harvesters/file/VaFileHarvester.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  protected val extractor = new VaFileExtractor()
+
+  /**
+    * Loads .zip files
+    *
+    * @param file File to parse
+    * @return TarInputstream of the tar contents
+    */
+  def getInputStream(file: File): Option[ZipInputStream] = {
```
mdellabitta: Because we're only handling zips in this class, and they should already be filtered by the FilenameFilter above, I wonder if you even need the match statement here.

AudreyAltman: I added a comment with this question so we don't lose it.  For now I'll leave it as is so its consistent with the other file harvesters.

PR #250, opened 2018-11-29
Mo harvester

AudreyAltman:
This adds a harvester and mapping for Missouri.  Both have been tested locally against a full MO data dump.  I left a few questions in the mapping `TODO` comments. 

The harvester includes logic to iterate through multiple zip files.  This isn't strictly necessary for MO since its zip contains only one file, but it keeps the MO harvester consistent with other harvesters.


AudreyAltman:
I added logic to the harvester to filter out deleted records.  It's a bit of a kludge but effective when tested it against the MO data dump.

PR #252, opened 2018-11-30
add entry to generate all reports

AudreyAltman:
This adds an entry to generate all reports (the same set of reports generated in `IngestRemap`).  I used this today to generate MO reports on EC2 with SBT.  The other entry point for reports, `ReporterMain.main`, is optimized to run one report at a time, rather than running all the reports at once.

### DIFF
--- a/src/main/scala/dpla/ingestion3/entries/ingest/ReportsEntry.scala
+++ b/src/main/scala/dpla/ingestion3/entries/ingest/ReportsEntry.scala
@@ -1,6 +1,6 @@
 package dpla.ingestion3.entries.ingest
 
-import dpla.ingestion3.confs.CmdArgs
+import dpla.ingestion3.confs.{CmdArgs, Ingestion3Conf, i3Conf}
 import dpla.ingestion3.entries.reports.ReporterMain.executeAllReports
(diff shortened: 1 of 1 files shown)


src/main/scala/dpla/ingestion3/entries/ingest/ReportsEntry.scala
```diff
@@ -0,0 +1,46 @@
...
+    val dataOut = cmdArgs.getOutput()
+    val shortName = cmdArgs.getProviderName()
+
+    // Get logger
+    val logger = Utils.createLogger("reports", shortName)
+
+    val sparkConf =
+      new SparkConf()
+        .setAppName("reports")
+        .setMaster("local[*]")
```
mdellabitta: I think there's a way to get the Spark master URL from the config file you might want to look at, because this won't use a cluster with `local[*]`

AudreyAltman: Good catch.  I pushed up a fix for this.  I actually copied this from the `JsonlEntry` so perhaps we should fix it there, too.

PR #254, opened 2018-12-06
delete unnecessary code

AudreyAltman:
This deletes an unnecessary and computationally expensive process in the EnrichExecutor.


AudreyAltman:
This is duplicated in #257

PR #256, opened 2018-12-11
tune mapping

AudreyAltman:
This optimizes mapping performance.

**Persistence**
Persistence has been kept minimal so as not to overload storage capacity, and tested for optimal performance.

**Accumulators and Counts**
I found that accumulators were more efficient than counts when working with very large DataFrames, and they helped minimize the number of data evaluations.  In some cases, the work of accumulators was being duplicated by counts.  In other cases, the same DataFrames were being counted repeatedly.  The `failureCount` accumulator was not being used.

**Repartitioning**
In the mapping, the input harvest data was repartitioned into 1024 segments.  This actually slowed the process once other adjustments had been made.  I repartitioned the `messages` DataFrame b/c it was lumpy.

The changes in `MessageProcessor` are inconsequential, just cleanup I did while reading the code.

### DIFF
--- a/src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
+++ b/src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
@@ -13,7 +13,6 @@ import org.apache.log4j.Logger
 import org.apache.spark.SparkConf
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.encoders.{ExpressionEncoder, RowEncoder}
-import org.apache.spark.sql.functions.col
 import org.apache.spark.storage.StorageLevel
 import org.apache.spark.util.LongAccumulator
 
@@ -170,7 +169,6 @@ trait MappingExecutor extends Serializable {
     // Transformation
     val messages: DataFrame = MessageProcessor.getAllMessages(results)(spark)
       .persist(StorageLevel.MEMORY_AND_DISK_SER)
-      .repartition(col("id"))
 
     val warnings: DataFrame = MessageProcessor.getWarnings(messages)
     val errors: DataFrame = MessageProcessor.getErrors(messages)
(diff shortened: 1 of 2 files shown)


src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
```diff
@@ -68,23 +66,24 @@ trait MappingExecutor extends Serializable {
     val oreAggregationEncoder: ExpressionEncoder[Row] = RowEncoder(model.sparkSchema)
 
     // Load the harvested record dataframe
-    val harvestedRecords: DataFrame = spark.read.avro(dataIn).repartition(1024)
+    val harvestedRecords: DataFrame = spark.read.avro(dataIn)
```
mdellabitta: IIRC I think this was in there to make Nara not be 1 partition based on the input data (sequencefiles built from their tar files). I think with our current harvester, which creates avros, it's totally unneeded.

src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
```diff
@@ -146,48 +159,48 @@ trait MappingExecutor extends Serializable {
...
 
-    // val results: Dataset[Row] = mappingResults.map(oreAgg => oreAgg)(oreAggregationEncoder)
-
-    val messages = MessageProcessor.getAllMessages(results)(spark)
-    val warnings = MessageProcessor.getWarnings(messages)
-    val errors =   MessageProcessor.getErrors(messages)
+    // Transformation
+    val messages: DataFrame = MessageProcessor.getAllMessages(results)(spark)
+      .persist(StorageLevel.MEMORY_AND_DISK_SER)
+      .repartition(col("id"))
```
mdellabitta: Are you sure about calling repartition() after persist()? Seems like you'd be repartitioning over and over potentially...

AudreyAltman: Agree, plus there are no discernible performance improvements.  Removed.

PR #257, opened 2018-12-11
Spark tuning - enrichment

AudreyAltman:
This optimized enrichment performance.

Much is similar to the mapping tuning in #256.  In addition, this eliminates proliferation of `EnrichmentDriver` instances, which has a significant impact on performance.

### DIFF
--- a/src/main/scala/dpla/ingestion3/executors/EnrichExecutor.scala
+++ b/src/main/scala/dpla/ingestion3/executors/EnrichExecutor.scala
@@ -74,6 +74,10 @@ trait EnrichExecutor extends Serializable {
     // Load the mapped records
     val mappedRows: DataFrame = spark.read.avro(dataIn)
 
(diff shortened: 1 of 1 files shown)


src/main/scala/dpla/ingestion3/executors/EnrichExecutor.scala
```diff
@@ -117,18 +126,34 @@ trait EnrichExecutor extends Serializable {
     // Write logs and summary reports.
     val endTime = System.currentTimeMillis()
 
+    // Get all the messages for all records.
+    val messages: DataFrame = MessageProcessor
+      .getAllMessages(successResults)(spark)
+      .persist(StorageLevel.MEMORY_AND_DISK_SER)
+      .repartition(col("id"))
```
mdellabitta: similar question about repartition after persist...

AudreyAltman: removed

PR #269, opened 2019-01-18
Date enrichment redux

moltude:
Rewrite of date enrichment. Only supports `yyyy` and `yyyy-yyyy` patterns. 
Add date reporting to enrichment summary and details enrichment reports.

### DIFF
--- a/src/main/scala/dpla/ingestion3/enrichments/date/DateBuilder.scala
+++ b/src/main/scala/dpla/ingestion3/enrichments/date/DateBuilder.scala
@@ -1,109 +1,53 @@
 package dpla.ingestion3.enrichments.date
 
+import dpla.ingestion3.model.DplaMapData.ZeroToOne
 import dpla.ingestion3.model._
-import scala.util.{Failure, Success, Try}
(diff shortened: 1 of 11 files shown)


src/main/scala/dpla/ingestion3/enrichments/date/DateBuilder.scala
```diff
@@ -1,109 +1,53 @@
...
-        simpleDateFormat
-      case Some(simpleDateFormat) => simpleDateFormat
+  /**
+    * Identifies an end date
+    *
+    * @param str Original date value
+    * @return Option[String]
+    */
+  def createEnd(str: String): Option[String] = str match {
+      case `str` if str.matches("[0-9]{4}") => Some(str)
```
mdellabitta: Won't this match both cases? You should probably either stick anchors in this regular expression for the start/end, or swap the order of these two.

mdellabitta: It's weird that this would pass the test, though. Am I wrong?

PR #272, opened 2019-01-25
Missouri mapping updates

moltude:
Add mappings for intermediateProvider, edmRights and type to Missouri mapping


src/main/scala/dpla/ingestion3/mappers/providers/MoMapping.scala
```diff
@@ -85,6 +92,9 @@ class MoMapping extends JsonMapping with JsonExtractor with IdMinter[JValue] wit
   override def title(data: Document[JValue]): AtLeastOne[String] =
     extractStrings(unwrap(data) \ "sourceResource" \ "title")
 
+  override def `type`(data: Document[JValue]): AtLeastOne[String] =
+    extractStrings(unwrap(data) \ "sourceResource" \ "type")
+
```
AudreyAltman: In the `Mapping` trait, the `type` method has a different return type:
```
def `type`(data: Document[T]): ZeroToMany[String] = Seq()
```
Do you want the return types to agree?

moltude: doh -- yes

PR #276, opened 2019-02-06
Duplicate original id error message 2

AudreyAltman:
A fresh PR, a fresh take on the `duplicateOriginalId` messaging.

@mdellabitta this is slightly different from what we discussed this morning.  This holds off on encoding the `OreAggregation`s altogether until after the new messages are added.  Let me know what you think.  

If you like this way of doing things, we could consider working with RDD[OreAggregation] for other functions too - such as filtering out successful results and building the final report - and then only do the encoding immediately before writes.  That could help prevent boo-boos like accidentally reordering column rows.

If you don't like this way of doing things, I'm ready to try something else.

I tested this locally by running `IngestRemap` over Oklahoma data.

### DIFF
--- a/src/main/resources/avro/MAPRecord.avsc
+++ b/src/main/resources/avro/MAPRecord.avsc
@@ -1080,11 +1080,10 @@
         "string"
       ]
     },
-  {
-    "name":
-    "messages",
-    "type": {
-      "name": "MessagesIngestMessageArray",
+    {
+      "name": "messages",
+      "type": {
+        "name": "MessagesIngestMessageArray",
         "type": "array",
         "items": {
           "name": "MessagesIngestMessage",
--- a/src/main/scala/dpla/ingestion3/messages/IngestMessageTemplates.scala
+++ b/src/main/scala/dpla/ingestion3/messages/IngestMessageTemplates.scala
@@ -56,6 +56,15 @@ trait IngestMessageTemplates {
       value = "both rights and edmRights are defined"
     )
 
+  def duplicateOriginalId(id: String): IngestMessage =
+    IngestMessage(
+      message = "Duplicate",
+      level = IngestLogLevel.warn,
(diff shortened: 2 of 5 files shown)


src/main/resources/avro/MAPRecord.avsc
```diff
@@ -1080,11 +1080,10 @@
         "string"
       ]
     },
-  {
-    "name":
-    "messages",
-    "type": {
-      "name": "MessagesIngestMessageArray",
```
AudreyAltman: This just fixes some spacing for legibility.

src/main/resources/avro/MAPRecord.avsc
```diff
@@ -1080,11 +1080,10 @@
...
+    {
+      "name": "messages",
+      "type": {
+        "name": "MessagesIngestMessageArray",
```
AudreyAltman: This just fixes some spacing for legibility.

mdellabitta: YOU MONSTER

src/main/scala/dpla/ingestion3/messages/IngestMessageTemplates.scala
```diff
@@ -56,6 +56,15 @@ trait IngestMessageTemplates {
       value = "both rights and edmRights are defined"
     )
 
+  def duplicateOriginalId(id: String): IngestMessage =
+    IngestMessage(
+      message = "Duplicate",
+      level = IngestLogLevel.warn,
```
moltude: My only suggestion is to change this from warning to error so that we fail all duplicate records rather than indexing one of them.

AudreyAltman: Will do, thanks @moltude

PR #279, opened 2019-02-27
Fix mapping double count

AudreyAltman:
**Problem**  
During mapping, the reported results for total attempted and total success were incorrect.  `mappingResults` was being evaluated twice, and so the accumulators that tracked these stats were double what they should be - and the mapping was unnecessarily inefficient / taking longer to run.  Also, total success was reported as equal to the total attempted even when not all records were successful.

**Solution**
1) Persist `mappingResults` so that it is only evaluated once.  Persisting this does add some memory pressure - I noticed for example that the Oklahoma mapping runs out of memory locally on the default memory setting - but we agreed that more memory pressure should be ok.

2) Get rid of the `totalCount` accumulator.  It was unnecessary because `encodedMappingResults.count` was already being evaluated and was equivalent to `totalCount`.

3) Get rid of the `successCount` accumulator.  It was incorrect, and accumulators can be unreliable, e.g. if nodes fail in the middle of computations.  Since @mdellabitta discovered that S3 has read-after-write consistency in our use case, we can use that instead to read in `successResults` and count them.  It is more reliable and adds very little time to the overall process.  Alternately, we could persist `successResults` so that we could write and count them without evaluating more than once, but that would add another dose of memory pressure - and it would probably overflow to disk so might not be much faster in a cluster environment than reading from S3.

**Testing**
I tested this locally with Oklahoma data.

BEFORE
```
                                Mapping Summary

Provider................................................................OKLAHOMA
Start date...................................................02/27/2019 12:44:49
Runtime.............................................................00:01:47.426

Attempted................................................................229,340
Successful...............................................................229,340
Failed.......................................................................213


                              Errors and Warnings

Messages
- Errors.....................................................................213
- Warnings...............................................................322,489

Records
- Errors.....................................................................213
- Warnings...............................................................111,943

                                Message Summary
Warnings
- Missing recommended field, language.....................................89,720
- Missing recommended field, publisher....................................51,307
- Missing recommended field, creator......................................47,890
- Missing recommended field, date.........................................43,869
- Missing recommended field, place........................................38,188
- Duplicate, rights and edmRights.........................................14,728
- Missing recommended field, type.........................................13,267
- Missing recommended field, format.......................................12,425
- Missing recommended field, description...................................9,951
- Missing recommended field, subject.......................................1,144
- Total..................................................................322,489
Errors
- Duplicate, originalId......................................................213
- Total......................................................................213
```

AFTER
```
                                Mapping Summary

Provider................................................................OKLAHOMA
Start date...................................................02/27/2019 14:56:21
Runtime.............................................................00:01:00.101

Attempted................................................................114,670
Successful...............................................................114,169
Failed.......................................................................213


                              Errors and Warnings

Messages
- Errors.....................................................................213
- Warnings...............................................................322,489

Records
- Errors.....................................................................213
- Warnings...............................................................111,943

                                Message Summary
Warnings
- Missing recommended field, language.....................................89,720
- Missing recommended field, publisher....................................51,307
- Missing recommended field, creator......................................47,890
- Missing recommended field, date.........................................43,869
- Missing recommended field, place........................................38,188
- Duplicate, rights and edmRights.........................................14,728
- Missing recommended field, type.........................................13,267
- Missing recommended field, format.......................................12,425
- Missing recommended field, description...................................9,951
- Missing recommended field, subject.......................................1,144
- Total..................................................................322,489
Errors
- Duplicate, originalId......................................................213
- Total......................................................................213
```

### DIFF
--- a/src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
+++ b/src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
@@ -54,8 +54,6 @@ trait MappingExecutor extends Serializable with IngestMessageTemplates {
       .config("spark.ui.showConsoleProgress", value = false)
       .getOrCreate()
 
-    val sc = spark.sparkContext
-
     // Need to keep this here despite what IntelliJ and Codacy say
     import spark.implicits._
 
@@ -219,7 +217,7 @@ trait MappingExecutor extends Serializable with IngestMessageTemplates {
         path
     }
 
-    val recordErrorCount: Long = MessageProcessor.getDistinctIdCount(errors)
+    val recordErrorCount: Long = attemptedCount - validRecordCount
     val recordWarnCount: Long = MessageProcessor.getDistinctIdCount(warnings)
 
     val errorMsgDetails: String =
--- a/src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
+++ b/src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
@@ -16,7 +16,6 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.encoders.{ExpressionEncoder, RowEncoder}
 import org.apache.spark.storage.StorageLevel
-import org.apache.spark.util.LongAccumulator
 
 import scala.collection.mutable
 import scala.util.{Failure, Success}
@@ -57,8 +56,6 @@ trait MappingExecutor extends Serializable with IngestMessageTemplates {
 
     val sc = spark.sparkContext
 
-    val successCount: LongAccumulator = sc.longAccumulator("Successful Record Count")
-
     // Need to keep this here despite what IntelliJ and Codacy say
     import spark.implicits._
 
@@ -76,7 +73,7 @@ trait MappingExecutor extends Serializable with IngestMessageTemplates {
       .select("document")
       .as[String]
       .rdd
-      .map(document => dplaMap.map(document, shortName, successCount))
+      .map(document => dplaMap.map(document, shortName))
       .persist(StorageLevel.MEMORY_AND_DISK_SER)
 
     // Get a list of originalIds that appear in more than one record
@@ -129,7 +126,7 @@ trait MappingExecutor extends Serializable with IngestMessageTemplates {
     successResults.write.avro(outputPath)
 
     // Get counts
-    val validRecordCount = successCount.count
+    val validRecordCount = spark.read.avro(outputPath).count // requires read-after-write consistency
     val attemptedCount = totalCount
 
     // Write manifest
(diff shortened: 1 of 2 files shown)


src/main/scala/dpla/ingestion3/executors/MappingExecutor.scala
```diff
@@ -57,28 +56,25 @@ trait MappingExecutor extends Serializable with IngestMessageTemplates {
...
     // Transformation only
-
-    val mappingResults: RDD[OreAggregation] = documents.rdd
-      .map(document => dplaMap.map(document, shortName, totalCount, successCount))
+    val mappingResults: RDD[OreAggregation] = harvestedRecords
+      .select("document")
+      .as[String]
+      .rdd
+      .map(document => dplaMap.map(document, shortName))
+      .persist(StorageLevel.MEMORY_AND_DISK_SER)
```
AudreyAltman: Lines 65-77 is just a refactor for clarity.

mdellabitta:
Q. Were the accumulators vastly incorrect once the `persist` call was put in? Is there a downside to keeping them? 

They do provide feedback during the course of the run in a way that `count` doesn't, and are viewable on the job dashboard... We could theoretically keep them in, but not use them for the final stats.

AudreyAltman:
@mdellabitta 
After the `persist` call was put in, the `totalCount` accumulator was correct, but the `successCount` accumulator was incorrect (i.e. it was equal to `totalCount` when it should have been less than `totalCount`).  

The only downside I can see to the accumulators is that they might be inaccurate for reasons beyond our control.  I don't think they add a lot of overhead, so I suppose we could keep them for dashboard-viewing purposes.

If we keep them, would you want to keep them both or just one so that we can monitor the mapping operation on the dashboard?  If we keep them both, we will need to do more troubleshooting to figure out why the `successCount` accumulator was incorrect.

AudreyAltman:
This last commit, https://github.com/dpla/ingestion3/pull/279/commits/f85979d51f169ec64fe2cf148b410b2c020fbc51, fixes inaccurate counts in the mapping summary.  

Before, the `Attempted` count did not equal sum of `Successful` and `Failed`.  The problem was that the `Failed` count was based on the unique record ids in all the error messages.  This count was inaccurate because some records have missing or duplicate ids.  Now, `Failed` is simply calculated by subtracting `Successful` from `Attempted`.

I also noticed that the `Error` counts in the `Message Summary` were inaccurate for `Duplicate, originalId` and `Missing required field, originalId`, and `Missing required field, dplaUri`.  This was because the error messages were being filtered to only take `distinct` messages.  In all three of these cases, identical errors can be created in different records.  Now, these counts are accurate.

src/main/scala/dpla/ingestion3/messages/MessageProcessor.scala
```diff
@@ -44,9 +44,29 @@ object MessageProcessor{
...
   }
 
-  def getErrors(ds: Dataset[Row]): Dataset[Row] =
-    ds.select("message", "level", "field", "id", "value")
-      .where(s"level=='${IngestLogLevel.error}'").distinct()
+  def getErrors(ds: Dataset[Row]): Dataset[Row] = {
+    val errors = ds.select("message", "level", "field", "id", "value")
+      .where(s"level=='${IngestLogLevel.error}'")
+
+    // TODO: Question for code review:
```
moltude: This assessment is correct, there was no need for distinct here.

AudreyAltman:
Removed `distinct` from `getErrors` method, since @moltude agreed it was unnecessary.

mdellabitta:
So glad you worked this out!

PR #283, opened 2019-03-14
LC CSV harvester

moltude:
Library of Congress CSV file harvester. Ingest LC data using the CSV output from the `ingestion3-api` harvester


src/main/scala/dpla/ingestion3/harvesters/file/LcCsvFileHarvester.scala
```diff
@@ -0,0 +1,152 @@
...
+          }
+        } match {
+          case Success(num) => num
+          case _ => 0
+        }
+      }
+    }
+  }
+
+  override def handleFile(fileResult: FileResult, unixEpoch: Long): Try[Int] = ???
```
moltude: @mdellabitta `handleCsv` replaces `handleFile` but I had to stub this method to extend `FileHarvster`.  I may have overlooked a way of doing this that conforms to what `handleFile` expects.

PR #284, opened 2019-03-26
Add Florida

moltude:
Add initial file harvester and mapping for Florida

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/FlMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/FlMapping.scala
@@ -119,7 +119,7 @@ class FlMapping extends JsonMapping with JsonExtractor with IngestMessageTemplat
   def extractDate(date: JValue): ZeroToMany[EdmTimeSpan] = {
     iterify(date).children.map(d =>
       EdmTimeSpan(
-        begin = extractString(d \ "start"),
+        begin = extractString(d \ "begin"),
         end = extractString(d \ "end"),
         originalSourceDate = extractString(d \ "displayDate")
       ))
--- a/src/main/scala/dpla/ingestion3/mappers/providers/FlMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/FlMapping.scala
@@ -16,33 +16,31 @@ class FlMapping extends JsonMapping with JsonExtractor with IngestMessageTemplat
   val formatBlockList: Set[String] = ExtentIdentificationList.termList
 
   // ID minting functions
-  // TODO: Should this be true or false?
-  override def useProviderName: Boolean = false
+  override def useProviderName: Boolean = true
 
-  // TODO: Should this be the same as provider short name?
-  override def getProviderName: String = "fl"
+  override def getProviderName: String = "florida"
 
   override def originalId(implicit data: Document[JValue]): ZeroToOne[String] =
     extractStrings(unwrap(data) \ "isShownAt").headOption
 
   // OreAggregation
-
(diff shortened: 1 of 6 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/FlMapping.scala
```diff
@@ -0,0 +1,127 @@
...
+  // OreAggregation
+  override def dataProvider(data: Document[JValue]): ZeroToMany[EdmAgent] =
+    extractStrings(unwrap(data) \ "dataProvider").map(nameOnlyAgent)
+
+  override def dplaUri(data: Document[JValue]): ZeroToOne[URI] = mintDplaItemUri(data)
+
+  override def edmRights(data: Document[json4s.JValue]): ZeroToMany[URI] =
+    extractStrings(unwrap(data) \ "sourceResource" \ "rights" \ "@id")
+      .filter(_.nonEmpty) // FIXME filtering non-empty values should be a standard edmRights normalization
+      .map(URI)
```
AudreyAltman: Calling out the `FIXME` in case you had intended to fix it for this PR

src/main/scala/dpla/ingestion3/mappers/providers/FlMapping.scala
```diff
@@ -0,0 +1,127 @@
...
+  override def extent(data: Document[JValue]): ZeroToMany[String] =
+    extractStrings(unwrap(data) \ "sourceResource" \ "extent")
+
+  override def format(data: Document[JValue]): ZeroToMany[String] =
+    extractStrings(unwrap(data) \ "sourceResource" \ "format") ++
+      extractStrings(unwrap(data) \ "sourceResource" \ "genre" \ "name")
+
+  override def genre(data: Document[JValue]): ZeroToMany[SkosConcept] =
+    extractStrings(unwrap(data) \ "sourceResource" \ "genre" \ "name")
+      .map(nameOnlyConcept)
```
AudreyAltman: Just sanity checking that Florida's `sourceResource.genre.name` is supposed to be mapped to both `format` and `genre` fields.

src/test/scala/dpla/ingestion3/mappers/providers/FlMappingTest.scala
```diff
@@ -0,0 +1,118 @@
...
+    )
+    assert(extractor.date(json) === expected)
+  }
+
+  // title
+  it should "extract the correct title" in {
+    val expected = List("The Ting Tings")
+    assert(extractor.title(json) === expected)
+  }
+}
```
AudreyAltman: You don't have tests for every field - is that intentional?

src/main/scala/dpla/ingestion3/mappers/providers/FlMapping.scala
moltude: Thanks -- I think that is for another PR. Just identifying some additional future work ;)

src/main/scala/dpla/ingestion3/mappers/providers/FlMapping.scala
moltude: That is what the mapping says. I think there may be some revisions once we inspect values in CQA but this is following the mapping to the letter.

src/test/scala/dpla/ingestion3/mappers/providers/FlMappingTest.scala
moltude: no. Let me fix this..

moltude:
Thanks -- re: file harvester abstractions - I totally agree but wanted to leave that for another PR and when we have a third JSON file extractor.

PR #286, opened 2019-03-27
TN Mapping

moltude:
Initial TN mapping and tests

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
@@ -19,11 +19,12 @@ class TnMapping extends XmlMapping with XmlExtractor with IngestMessageTemplates
       FormatTypeValuesBlockList.termList
 
   // ID minting functions
-  override def useProviderName(): Boolean = true
+  override def useProviderName(): Boolean = false
 
   override def getProviderName(): String = "tn"
 
   override def originalId(implicit data: Document[NodeSeq]): ZeroToOne[String] =
+  // ingestion1 ID minting >> /select-id?prop=id&use_source=no
     extractString(data \ "header" \ "identifier")
       .map(_.trim)
 
@@ -241,6 +242,6 @@ class TnMapping extends XmlMapping with XmlExtractor with IngestMessageTemplates
   // Helper method
   def agent = EdmAgent(
     name = Some("Tennessee Digital Library"),
-    uri = Some(URI("http://dp.la/api/contributor/tn"))
+    uri = Some(URI("http://dp.la/api/contributor/tennessee"))
   )
 }
--- a/src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
@@ -73,7 +73,7 @@ class TnMapping extends XmlMapping with XmlExtractor with IngestMessageTemplates
 
   override def description(data: Document[NodeSeq]): Seq[String] =
   // <mods:abstract>
-    extractStrings(data \\ "mods" \ "abstract")
+    extractStrings(data \\ "metadata" \ "mods" \ "abstract")
 
   override def extent(data: Document[NodeSeq]): ZeroToMany[String] =
   // <mods:physicalDescription><mods:extent>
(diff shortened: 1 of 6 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
```diff
@@ -0,0 +1,265 @@
...
+      .filter(node => (node \ "@type").text.equalsIgnoreCase("references"))
+      .flatMap(node =>
+        extractStrings(node \ "title" \ "titlePart") ++
+        extractStrings(node \ "location" \ "url"))
+
+  override def rights(data: Document[NodeSeq]): AtLeastOne[String] =
+    // <mods:accessCondition>
+    (data \\ "mods" \ "accessCondition")
+      .flatMap(node => getByAttribute(node, "type", "local rights statement")) // TODO should this filter be in place?
+      .flatMap(extractStrings)
```
AudreyAltman: Calling out the `TODO` above in case you had intended to address it in this PR.

src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
```diff
@@ -0,0 +1,265 @@
...
+
+  override def temporal(data: Document[NodeSeq]): ZeroToMany[EdmTimeSpan] =
+  // <mods:subject><mods:temporal>
+    extractStrings(data \\ "mods" \ "subject" \ "temporal")
+      .map(stringOnlyTimeSpan)
+
+  override def title(data: Document[NodeSeq]): Seq[String] =
+  // <mods:titleInfo><mods:title>
+    (data \ "metadata" \ "mods" \ "titleInfo" \ "title")
+      .filter(node => node.attributes.isEmpty)
```
AudreyAltman: I see that you don't want to capture attributes like `type=alternative`. Just double-checking that there is no possibility that the title could have different attribute but would still qualify as a title.

src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
```diff
@@ -0,0 +1,265 @@
...
+  override def dataProvider(data: Document[NodeSeq]): ZeroToMany[EdmAgent] =
+  // <recordInfo><recordContentSource>
+    (data \\ "mods" \ "recordInfo" \ "recordContentSource")
+      .flatMap(extractStrings)
+      .map(nameOnlyAgent)
+
+  override def edmRights(data: Document[NodeSeq]): ZeroToMany[URI] =
+  // <accessCondition type="use and reproduction" xlink="[value to be mapped is here]">
+    (data \ "metadata" \ "mods" \ "accessCondition")
+      .flatMap(node => getByAttribute(node, "type", "use and reproduction"))
```
AudreyAltman: Any need to be case-insensitive for matching `"use and reproduction"`?

src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
```diff
@@ -0,0 +1,265 @@
...
+      .flatMap(node => node.attribute(node.getNamespace("xlink"), "href"))
+      .flatMap(extractStrings(_))
+      .map(URI)
+
+  override def intermediateProvider(data: Document[NodeSeq]): ZeroToOne[EdmAgent] =
+    // <note displayLabel=“Intermediate Provider”>
+    (data \\ "mods" \ "note")
+      .flatMap(node => getByAttribute(node, "displayLabel", "Intermediate Provider"))
```
AudreyAltman: Any need to be case-insensitive for matching `"Intermediate Provider"`?

src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
```diff
@@ -0,0 +1,265 @@
...
+      .flatMap(extractStrings)
+      .map(nameOnlyAgent)
+      .headOption
+
+  override def isShownAt(data: Document[NodeSeq]): ZeroToMany[EdmWebResource] =
+  // <location><url usage = starts with "primary" access="object in context">
+    (data \\ "mods" \ "location" \ "url")
+      .filter(node => (node \ "@usage").text.startsWith("primary"))
+      .flatMap(node => getByAttribute(node, "access", "object in context"))
```
AudreyAltman: More sanity checking about case-insensitivity.

src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
```diff
@@ -0,0 +1,265 @@
...
+      .flatMap(extractStrings)
+      .map(stringOnlyWebResource)
+
+  override def `object`(data: Document[NodeSeq]): ZeroToMany[EdmWebResource] =
+    // <location><url access="raw object">
+    (data \\ "mods" \ "location" \ "url")
+      .flatMap(node => getByAttribute(node.asInstanceOf[Elem], "access", "raw object"))
+      .flatMap(extractStrings)
```
AudreyAltman: Still more sanity checking about case insensitivity.

src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
```diff
@@ -0,0 +1,265 @@
...
+      .map(stringOnlyWebResource)
+
+  override def originalRecord(data: Document[NodeSeq]): ExactlyOne[String] = Utils.formatXml(data)
+
+  override def preview(data: Document[NodeSeq]): ZeroToMany[EdmWebResource] =
+  // <location><url access="preview">
+    (data \\ "mods" \ "location" \ "url")
+      .flatMap(node => getByAttribute(node.asInstanceOf[Elem], "access", "preview"))
```
AudreyAltman: Case insensitivity?

src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
moltude: That was for partner follow-up in CQA.

moltude:
There is a fix for case-insensitive matching for `getByAttribute` in the `Extractor` class in this PR.

src/main/scala/dpla/ingestion3/mappers/providers/TnMapping.scala
moltude: There were only 58 records missing a title and I believe all them are legitimately missing any title property.

PR #287, opened 2019-03-27
Illinois mapping 

moltude:
Add Illinois mapping and tests. 

ID minting will need to change but wanted to implement with consistent IDs for now to facilitate record comparison between CQA and production portal. Collection mapping is also problematic and needs discussion with hub for resolution.

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/IllinoisMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/IllinoisMapping.scala
@@ -19,8 +19,6 @@ class IllinoisMapping extends XmlMapping with XmlExtractor with IngestMessageTem
     DigitalSurrogateBlockList.termList ++
       ExtentIdentificationList.termList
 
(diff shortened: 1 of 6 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/IllinoisMapping.scala
```diff
@@ -0,0 +1,134 @@
...
+    extractStrings(data \\ "metadata" \\ "alternative")
+
+  override def contributor(data: Document[NodeSeq]): ZeroToMany[EdmAgent] =
+    // contributor
+    extractStrings(data \\ "metadata" \\ "contributor")
+      .map(nameOnlyAgent)
+
+  override def creator(data: Document[NodeSeq]): ZeroToMany[EdmAgent] =
+    // contributor
+    extractStrings(data \\ "creator")
```
AudreyAltman: I think the comment should say `// creator`

PR #290, opened 2019-04-02
first draft IN mapping

AudreyAltman:
This is a first pass at the Indiana mapping.

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/InMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/InMapping.scala
@@ -1,6 +1,7 @@
 package dpla.ingestion3.mappers.providers
 
 import dpla.ingestion3.enrichments.normalizations.StringNormalizationUtils._
+import dpla.ingestion3.enrichments.normalizations.filters.DigitalSurrogateBlockList
 import dpla.ingestion3.mappers.utils.{Document, XmlExtractor, XmlMapping}
 import dpla.ingestion3.messages.IngestMessageTemplates
 import dpla.ingestion3.model.DplaMapData._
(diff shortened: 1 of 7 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/InMapping.scala
```diff
@@ -0,0 +1,140 @@
...
+  override def originalId(implicit data: Document[NodeSeq]): ZeroToOne[String] =
+    extractString(data \ "header" \ "identifier")
+
+  override def dplaUri(data: Document[NodeSeq]): ZeroToOne[URI] = mintDplaItemUri(data)
+
+  override def dataProvider(data: Document[NodeSeq]): ZeroToMany[EdmAgent] =
+    extractStrings(data \ "metadata" \ "qualifieddc" \ "provenance")
+      .map(nameOnlyAgent)
+
+  // TODO: In ingest1, the following are also mapped to edmRights:
```
moltude: @AudreyAltman yes, I think that is a valid assumption

src/main/scala/dpla/ingestion3/mappers/providers/InMapping.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  override def extent(data: Document[NodeSeq]): ZeroToMany[String] =
+    extractStrings(data \ "metadata" \ "qualifieddc" \ "extent")
+
+  override def format(data: Document[NodeSeq]): ZeroToMany[String] =
+    extractStrings(data \ "metadata" \ "qualifieddc" \ "medium")
+      .flatMap(_.splitAtDelimiter(";"))
+
+  // TODO: This is in the crosswalk but not in ingest1. It is also identical to format mapping.  Keep?
+  override def genre(data: Document[NodeSeq]): ZeroToMany[SkosConcept] =
```
moltude: genre is a deprecated field in MAPv5 so this is going to be dropped in the near term anyway. I'm 50/50 on keeping it [tilt towards removing] as I do not believe `genre` is made available in the API.

src/main/scala/dpla/ingestion3/mappers/providers/InMapping.scala
```diff
@@ -0,0 +1,140 @@
...
+      .map(stringOnlyTimeSpan)
+
+  override def title(data: Document[NodeSeq]): AtLeastOne[String] =
+    extractStrings(data \ "metadata" \ "qualifieddc" \ "title")
+
+  // TODO: Watch out for this field during QA.
+  //  Ingest1 has the following enrichments that may indicate the need for some logic here:
+  //  "/enrich-type?default=image" ?
+  //  "/enrich-type?send_rejects_to_format=true"
+  override def `type`(data: Document[NodeSeq]): ZeroToMany[String] =
```
moltude: good catch. The solution might be to map the type field to format and then apply a format allowFilter on the type values. 

Not sure if we want to apply a default type value though...that seems awkward.

AudreyAltman:
@moltude I've addressed your comments here: https://github.com/dpla/ingestion3/pull/290/commits/f3dd5d98507f63da6ff5ccc8e3e88a236466ab3b Could you review them and make sure it all looks good?

AudreyAltman:
@mdellabitta strip trailing semicolons from `edmRights` is implemented here: https://github.com/dpla/ingestion3/pull/290/commits/febb65dbac3aee7acb46e2ecfb0b652e17c6476c

PR #291, opened 2019-04-10
Feature/smithsonian

moltude:
Add Smithsonian XML file harvester
Add Smithsonian mapping & tests 
Update string type -> DcmiType mapping with SI specific terminology 
Add SI to ProviderProfiles and ProviderRegistry


src/main/scala/dpla/ingestion3/enrichments/TypeEnrichment.scala
```diff
@@ -15,7 +15,6 @@ class TypeEnrichment extends VocabEnrichment[String] {
   val dcmiType = DCMIType()
 
   val DcmiTypeMap: Map[String, IRI] = Map(
-    "appliance" -> dcmiType.Image,
```
mdellabitta: Umm, what about my collection of washer/dryer photos?

AudreyAltman: @mdellabitta and the dad jokes strike again :)

mdellabitta: Damn straight!

moltude: laugh but I spent more time with this book in my public history class than I care to remember 
`https://www.amazon.com/Save-Womens-Lives-History-Machines/dp/0972971009`

PR #295, opened 2019-05-03
NARA incremental updates

moltude:
Update NARA harvester to work with incremental updates 

Requires .tar.gz files to be named in a consistent way so they can be sorted chronologically [e.g. 2018105-data1.tar.gz, 20190425-delta1.tar.gz]
After all records are processed records are grouped by ID and sorted using their originating tar.gz filename. Only the most recent version of an item is harvested.

Deleted supported by including a `deletes.xml` file at the root level of the input directory


moltude:
@mdellabitta  [sample test data](https://drive.google.com/drive/folders/1zxq9klj0_DvaozZC3Wh4oU_I8Wq78ejk?usp=sharing)

PR #299, opened 2019-05-20
add MWDL format mapping

AudreyAltman:
Add format mapping for MWDL.
Should we apply any filters now, or wait until the next ingest cycle and do CQA?


AudreyAltman:
@moltude thanks, I've added the format filter

AudreyAltman:
@moltude I've updated the `isShownAt` URL as well.

PR #301, opened 2019-05-24
Maryland migration

AudreyAltman:
This adds a mapping for Maryland, migrated from ingestion1.  The mapping has been tested thoroughly with automated reports.  A property value report for `preview` is added to assist with automated testing. The property value report for `isShownAt` is fixed.

One significant change from ingestion1 is that there is no collection mapping.  This is because there doesn't appear to be collection data Maryland's records anymore.  None of the harvested records contained any collection data.

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/MarylandMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/MarylandMapping.scala
@@ -1,7 +1,7 @@
 package dpla.ingestion3.mappers.providers
 
 import dpla.ingestion3.enrichments.normalizations.StringNormalizationUtils._
-import dpla.ingestion3.mappers.utils.{Document, XmlMapping, XmlExtractor}
+import dpla.ingestion3.mappers.utils.{Document, XmlExtractor, XmlMapping}
 import dpla.ingestion3.model.DplaMapData._
 import dpla.ingestion3.model._
 import dpla.ingestion3.utils.Utils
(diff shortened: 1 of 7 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/MarylandMapping.scala
```diff
@@ -0,0 +1,125 @@
...
+      .flatMap(_.splitAtDelimiter(";"))
+
+  // OreAggregation
+  override def dplaUri(data: Document[NodeSeq]): ZeroToOne[URI] = mintDplaItemUri(data)
+
+  override def dataProvider(data: Document[NodeSeq]): ZeroToMany[EdmAgent] =
+    extractStrings(data \ "metadata" \\ "source")
+      .flatMap(_.splitAtDelimiter(";"))
+      .map(nameOnlyAgent)
+      .slice(0, 1) // get first instance
```
mdellabitta: Would `head` or `headOption` make more sense?

mdellabitta: This probably doesn't matter at all, just thinking out loud.

AudreyAltman: Yeah, I just did it that way b/c `slice` returns an empty array if the value is missing, so you don't have to do another step to get the correct return type.

mdellabitta: TIL. That's awesome!

PR #313, opened 2019-07-10
Add DLG file harvester

moltude:
Update DLG provider profile
Remove old DLG API harvester

Data to use is in `s3a://dpla-hub-ga/06102019/`


src/main/scala/dpla/ingestion3/harvesters/file/DlgFileHarvester.scala
```diff
@@ -0,0 +1,154 @@
...
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.{DataFrame, SparkSession}
+import org.json4s.jackson.JsonMethods._
+import org.json4s.{JValue, _}
+
+import scala.util.{Failure, Success, Try}
+
+
+/**
+  * Entry for performing a Florida file harvest
```
mdellabitta: Should say DLG

src/main/scala/dpla/ingestion3/harvesters/file/DlgFileHarvester.scala
```diff
@@ -0,0 +1,154 @@
...
+  */
+class DlgFileHarvester(spark: SparkSession,
+                       shortName: String,
+                       conf: i3Conf,
+                       logger: Logger)
+  extends FileHarvester(spark, shortName, conf, logger) {
+
+  def mimeType: String = "application_json"
+
+  protected val extractor = new FlFileExtractor()
```
mdellabitta: Should there be a separate implementation of this that's for DLG, or should we generify the name of this somehow?

moltude: just renamed for now, should be a part of the larger file harvester refactor though. Probably as a generic XmlExtractor and JsonExtractor rather than provider specific

src/main/scala/dpla/ingestion3/harvesters/file/DlgFileHarvester.scala
moltude: done

PR #314, opened 2019-07-15
Feature/digitalnc mapper

moltude:


src/main/scala/dpla/ingestion3/mappers/providers/NcMapping.scala
```diff
@@ -0,0 +1,190 @@
...
+
+  override def rights(data: Document[NodeSeq]): AtLeastOne[String] =
+  // <accessCondition type="local rights statements">
+    (data \\ "mods" \ "accessCondition")
+      .filter(node => filterAttribute(node, "type", "local rights statements"))
+      .flatMap(extractStrings)
+
+  override def subject(data: Document[NodeSeq]): Seq[SkosConcept] =
+    // <mods:subject><mods:topic>
+    extractStrings(data \\ "mods" \ "subject" \ "topic").map(nameOnlyConcept)
```
AudreyAltman: For consistency, you could put a line break before `.map(nameOnlyConcept)`

src/main/scala/dpla/ingestion3/mappers/providers/NcMapping.scala
```diff
@@ -0,0 +1,190 @@
...
+  override def intermediateProvider(data: Document[NodeSeq]): ZeroToOne[EdmAgent] = {
+    // second <note type=ownership> if it exists
+    val providers = (data \\ "mods" \ "note")
+      .filter(node => filterAttribute(node, "type", "ownership"))
+      .flatMap(extractStrings)
+      .map(nameOnlyAgent)
+
+    if (providers.length > 1)
+      Some(providers(1))
+    else None
```
AudreyAltman: I think you could do `.lift(1)` instead of the `if-else` here.  It will return an `Option` with `None` if there is no element at index 1.

src/main/scala/dpla/ingestion3/mappers/providers/NcMapping.scala
```diff
@@ -0,0 +1,190 @@
...
+    extractString(data \ "header" \ "identifier")
+      .map(_.trim)
+
+  // SourceResource mapping
+//  override def collection(data: Document[NodeSeq]): ZeroToMany[DcmiTypeCollection] =
+//  // FIXME should pull from setSpec
+//    (data \\ "mods" \ "relatedItem")
+//      .flatMap(node => getByAttribute(node.asInstanceOf[Elem], "type", "host"))
+//      .flatMap(collection => extractStrings(collection \ "titleInfo" \ "title"))
+//      .map(nameOnlyCollection)
```
AudreyAltman: Did you want to leave this commented-out method here?  Ok either way, just checking.

src/main/scala/dpla/ingestion3/mappers/providers/NcMapping.scala
moltude: 👍 thanks for the reminder

src/main/scala/dpla/ingestion3/mappers/providers/NcMapping.scala
moltude: done

src/main/scala/dpla/ingestion3/mappers/providers/NcMapping.scala
moltude: Going to leave it commented out for now. Added some additional comments explaining why it is there for now.

PR #315, opened 2019-07-17
Hathi mapper

AudreyAltman:
This adds a MARC-XML mapping for Hathi.

The preview mapping is the strangest addition - a separate HTTP request must be made for every record in order to retrieve the preview URL.  This slows mapping, but is manageable.  On 4-worker cluster, the mapping ran in just under three hours.

**One change that will affect all mappings going forward:**
Hathi had a very large amount of exact duplicate records in the harvest, which significantly slowed runtime for the mapping and caused many records to fail mapping because of duplicate original IDs.  This filters out harvested records that are exact duplicates before the mapping runs.  It reports the number of duplicates in the mapping summary.  Please let me know if this will cause problems for other mappings.

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/HathiMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/HathiMapping.scala
@@ -1,0 +1,699 @@
+package dpla.ingestion3.mappers.providers
+
+import java.net.URL
+
+import dpla.ingestion3.enrichments.normalizations.StringNormalizationUtils._
+import dpla.ingestion3.mappers.utils.{Document, JsonExtractor, XmlExtractor, XmlMapping}
+import dpla.ingestion3.model.DplaMapData._
+import dpla.ingestion3.model._
+import dpla.ingestion3.utils.{HttpUtils, Utils}
+import org.json4s.jackson.JsonMethods._
+import org.json4s.JValue
+import org.json4s.JsonDSL._
+
+import scala.annotation.tailrec
+import scala.util.{Failure, Success, Try}
+import scala.xml._
+
+
+class HathiMapping extends XmlMapping with XmlExtractor {
+
+  val isShownAtPrefix: String = "http://catalog.hathitrust.org/Record/"
+
+  // ID minting functions
+  override def useProviderName: Boolean = true
+
+  override def getProviderName: String = "hathitrust"
+
+  override def originalId(implicit data: Document[NodeSeq]): ZeroToOne[String] =
+    // <controlfield> tag = 001
+    controlfield(data, Seq("001"))
+      .flatMap(extractStrings)
+      .headOption
+
+  // SourceResource mapping
+
+  override def contributor(data: Document[NodeSeq]): ZeroToMany[EdmAgent] =
+   // <datafield> tag = 700, 710, 711, or 720
+    marcFields(data, Seq("700", "710", "711", "720"))
+      .filterNot(filterSubfields(_, Seq("e")) // exclude subfields with @code=e and...
+      .flatMap(extractStrings)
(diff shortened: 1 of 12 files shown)


src/main/scala/dpla/ingestion3/mappers/utils/Extractor.scala
```diff
@@ -91,6 +91,17 @@ trait XmlExtractor extends Extractor[NodeSeq] {
     getByAttribute(e.asInstanceOf[Elem], att, value)
   }
 
+  def filterAttributeListOptions(node: Node, att: String, values: Seq[String]): Boolean =
```
moltude: Can you write a quick definition of these methods? E.g. given X what do they return?

src/main/scala/dpla/ingestion3/mappers/providers/HathiMapping.scala
```diff
@@ -0,0 +1,703 @@
...
+    "uiug" -> "University of Illinois",
+    "uiuo" -> "University of Illinois",
+    "umn" -> "University of Minnesota",
+    "usu" -> "Utah State University Press",
+    "uva" -> "University of Virginia",
+    "wu" -> "University of Wisconsin",
+    "yale" -> "Yale University"
+  )
+
+  private val rightsMapping: Map[String, String] = Map(
```
mdellabitta: Seems like there's a pretty easy rightsstatements implementation here. Do most of the values in their plain-text rights statements map to one of these?

src/main/scala/dpla/ingestion3/mappers/providers/HathiMapping.scala
```diff
@@ -0,0 +1,703 @@
...
+    if (hathiId.isEmpty || oclcId.isEmpty || googlePrefix.isEmpty) None
+    else {
+      val isbnSuffix = if (isbnId.isEmpty) "" else ",ISBN:" + isbnId
+      val url = baseUrl + googlePrefix + ":" + hathiId + ",OCLC:" + oclcId + isbnSuffix
+      Some(url)
+    }
+  }
+
+  // Control flow for entire process of constructing request URL, sending request, and processing response.
+  val thumbnailUrl: Option[String] = {
```
mdellabitta: It might be cleaner to chain calls to `.flatMap` for this.

AudreyAltman: good idea

src/main/scala/dpla/ingestion3/mappers/providers/HathiMapping.scala
```diff
@@ -0,0 +1,703 @@
...
+        case _ => Seq()
+      }
+
+      cDate.map(stringOnlyTimeSpan)
+    }
+  }
+
+  override def description(data: Document[NodeSeq]): ZeroToMany[String] =
+    // <datafield> tag = any number in the 500s except 538
+    marcFields(data, (500 to 599).filterNot(_ == 538).map(_.toString))
```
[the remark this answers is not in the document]
AudreyAltman: Haha that is a good suggestion.  I went halfway and pulled out range as an immutable value, so it only has to be computed once.  I had something similar for `subject` so I just did the same thing here.

src/main/scala/dpla/ingestion3/mappers/utils/Extractor.scala
AudreyAltman: Yes, thank you for reminding me!

src/main/scala/dpla/ingestion3/mappers/providers/HathiMapping.scala
AudreyAltman: Yes, Scott and I made a ticket for this - Scott suggested that we should check with Hathi first to see what version of the licenses they are using.

PR #316, opened 2019-07-18
Add Harvard

moltude:
Add Harvard mapping and tests
Add term to type enrichment mapping

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/HarvardMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/HarvardMapping.scala
@@ -48,21 +48,20 @@ class HarvardMapping extends XmlMapping with XmlExtractor with IngestMessageTemp
   override def extent(data: Document[NodeSeq]): ZeroToMany[String] =
     extractStrings(data \ "metadata" \ "mods" \ "physicalDescription" \ "extent")
 
-  override def format(data: Document[NodeSeq]): ZeroToMany[String] =
-    extractStrings(data \ "metadata" \ "mods" \ "genre")
-      .map(
-        _.applyBlockFilter(
-          DigitalSurrogateBlockList.termList ++
-            ExtentIdentificationList.termList
-        )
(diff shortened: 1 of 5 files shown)


moltude:
screenshot-2014-07-22-22-12-26

src/main/scala/dpla/ingestion3/mappers/providers/HarvardMapping.scala
```diff
@@ -0,0 +1,377 @@
...
+      } yield child.text.trim
+
+      if (candidates.isEmpty) None else Some(candidates.mkString(" ").trim)
+
+    case _ => None
+  }
+
+
+  //Helper method to get a list of creators and contributors
+  private def processNames(data: Document[NodeSeq]): Names = {
```
AudreyAltman: I think of you made this a `val` instead of a `def` it would only execute once per mapping.  Probably not a big deal since Harvard is small.

src/main/scala/dpla/ingestion3/mappers/providers/HarvardMapping.scala
```diff
@@ -0,0 +1,377 @@
...
+
+    val titleSubjects = for {
+      subjectNode <- data \ "metadata" \ "mods" \ "subject" \ "titleInfo"
+      subjectText <- processTitleInfo(subjectNode)
+    } yield nameOnlyConcept(subjectText)
+
+    topicSubjects ++ nameSubjects ++ titleSubjects
+  }
+
+  //todo
```
AudreyAltman: delete `//todo` comment?

moltude: done

src/main/scala/dpla/ingestion3/mappers/providers/HarvardMapping.scala
moltude: done

PR #317, opened 2019-08-09
Lda initial

AudreyAltman:
This adds a process that to produce LDA vectors for each enriched record.  This process stands alone from other ingest processes.  The invocation is a bit verbose and is documented in `TopicModelEntry`.  It takes enriched records (read from an avro file) and writes out a dataframe with the record ID, the LDA vector, and a few other artifacts such as the lemmas.  

The process of creating and training the LDA model is separate. This code takes new, unseen records and processes them through an existing LDA model.  We do _not_ need to retrain the LDA model every time.

There are three main processes:
1) Lemmatize the subject, title, and description words
2) Create a "bag of words" by filtering out stopwords, lowercasing, etc.
3) Get the topic distribution vector for each document

I ran this locally on OK hub and it took about 17 mins.  If you want to try running it, I can give you the s3 address for the additional inputs (LDA model, stopwords file, etc.).  

I took extraneous parenthesis out of methods having to do with getting command line arguments b/c my compiler was fussing about it. This was just a bit of cleanup.

I do not know what to do about the new security vulnerability - any help/advice would be appreciated!


AudreyAltman:
This build works locally with SBT.  I am still unsure about whether or not this will work with `spark-submit`

PR #326, opened 2019-10-09
MDL harvester rewrite

moltude:


src/main/scala/dpla/ingestion3/harvesters/api/MdlHarvester.scala
```diff
@@ -95,18 +102,40 @@ class MdlHarvester(spark: SparkSession,
...
+  def getFirstUrl(queryParams: Map[String, String]): String = {
+    val url = new URIBuilder()
+      .setScheme("https")
+      .setHost("lib-metl-prd-01.oit.umn.edu")
+      .setPath("/api/v2/records")
+      .setParameter("page[size]", queryParams.getOrElse("rows", defaultRows))
+
+
+    /**
+      * This is a real mangling of shit
```
mdellabitta: :hankey:

src/main/scala/dpla/ingestion3/harvesters/api/MdlHarvester.scala
```diff
@@ -95,18 +102,40 @@ class MdlHarvester(spark: SparkSession,
...
+        new Parameters(query_tuple(0), query_tuple(1))
+      })
+    // Add query parameters to URL
+    params.foreach(p => url.addParameter(p.getName, p.getValue))
+
+    url.build().toURL.toString
+  }
 }
+
+class Parameters(name: String, value: String) {
```
mdellabitta: case class? then you don't need the accessor methods.

moltude:
I don't think I'm going to pick this up again until they've taken a swing at optimizing the API on their end and have a firmer idea of what might change again.

src/main/scala/dpla/ingestion3/harvesters/api/MdlHarvester.scala
moltude: done

PR #330, opened 2019-10-16
Correctly hangle last page of harvest from MDL API

moltude:
The last page of records in a harvest will not have a `links \ next` property. Therefore, change `requestUrl` to `Option[String]` and remove unnecessary `continueHarvest` var.

### DIFF
--- a/src/main/scala/dpla/ingestion3/harvesters/api/MdlHarvester.scala
+++ b/src/main/scala/dpla/ingestion3/harvesters/api/MdlHarvester.scala
@@ -33,47 +33,40 @@ class MdlHarvester(spark: SparkSession,
   override def localHarvest: DataFrame = {
     implicit val formats = DefaultFormats
 
-    // Mutable vars for controlling harvest loop
-    var continueHarvest = true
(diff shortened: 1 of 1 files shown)


src/main/scala/dpla/ingestion3/harvesters/api/MdlHarvester.scala
```diff
@@ -33,47 +33,40 @@ class MdlHarvester(spark: SparkSession,
...
     implicit val formats = DefaultFormats
 
-    // Mutable vars for controlling harvest loop
-    var continueHarvest = true
-    var requestUrl = getFirstUrl(queryParams)
+    // Mutable var for controlling harvest loop
+    var requestUrl: Option[String] = Some(getFirstUrl(queryParams))
 
-    while (continueHarvest) getSinglePage(requestUrl) match {
+    while (requestUrl.isDefined) getSinglePage(requestUrl.get) match {
```
mdellabitta: If you remove the call to `.get` and pattern match on Some(x) etc, you'll get rid of a warning about the unsafe call to get.

PR #335, opened 2019-10-22
Add David Rumsey

moltude:
Add David Rumsey mapping and associated tests.

FYI -- this mapping will break/change Rumsey IDs


src/main/scala/dpla/ingestion3/mappers/providers/RumseyMapping.scala
```diff
@@ -0,0 +1,154 @@
...
+  // OreAggregation
+  override def dplaUri(data: Document[NodeSeq]): ZeroToOne[URI] = mintDplaItemUri(data)
+
+  // done, sbw hard-coded
+  override def dataProvider(data: Document[NodeSeq]): ZeroToMany[EdmAgent] =
+    Seq(EdmAgent(name = Some("David Rumsey")))
+
+  // done, sbw hard-coded
+  override def edmRights(data: Document[NodeSeq]): ZeroToMany[URI] =
+    Seq(URI("https://creativecommons.org/licenses/by-nc-sa/3.0/"))
```
mdellabitta: We should probably set up constants for these URIs. The canonical URI is `http`, not `https`

src/main/scala/dpla/ingestion3/mappers/providers/RumseyMapping.scala
```diff
@@ -0,0 +1,154 @@
...
+
+  // take the first URL from dc:identifier
+  override def isShownAt(data: Document[NodeSeq]): ZeroToMany[EdmWebResource] =
+    extractStrings(metadataRoot(data) \ "identifier")
+      .filter(Utils.isUrl)
+      .take(1)
+      .map(_.replace("74.126.224.22", "www.davidrumsey.com")) // replace hard coded IP address in value
```
mdellabitta: :cry:

src/main/scala/dpla/ingestion3/mappers/providers/RumseyMapping.scala
moltude: @mdellabitta Did I/we muck the edmRights normalization? 

https://github.com/dpla/ingestion3/blob/5eb531d24586ab7bbeb7239a93b2ec903fd3d39e/src/main/scala/dpla/ingestion3/enrichments/normalizations/StringNormalizations.scala#L76-L85

mdellabitta: Looks like it: https://rightsstatements.org/page/1.0/?language=en#collection-nc

D'oh!

PR #340, opened 2019-12-10
Fix for Texas harvest / oai set blacklist

mdellabitta:
- Removes old oai harvester.

- Rearranges BlacklistOaiRelation so the sets are decided locally 
and then parallelized, which seems to clean up the execution plan.


moltude:
Premature bucket cleaning....📏 ✋

PR #345, opened 2019-12-17
Feature/validate mapped fields only

moltude:
Adds logic to conditionally validate recommended but optional fields in DPLA MAP. 
* creator
* date
* format
* language
* place
* publisher
* subject
* type

By default all recommended fields are checked for the presence of at least one value and if they do not have at least one value a warning message is logged. If a data provider does provide a mapping for one of these fields, the mapping should override the `enforce[Field]` value and set it to false. 

`override val enforceCreator = false // no mapping for creator provided, do not validate`

### DIFF
--- a/src/main/scala/dpla/ingestion3/model/package.scala
+++ b/src/main/scala/dpla/ingestion3/model/package.scala
@@ -57,7 +57,7 @@ package object model {
 
   def emptyEdmWebResource: EdmWebResource = stringOnlyWebResource("")
 
-  def emptyJValue: JValue = "" -> ""
+  def emptyJValue: JValue = JNothing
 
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/model/package.scala
```diff
@@ -57,6 +57,8 @@ package object model {
 
   def emptyEdmWebResource: EdmWebResource = stringOnlyWebResource("")
 
+  def emptyJValue: JValue = "" -> ""
```
mdellabitta: Wondering out loud: Would JNothing or JNull work here?

moltude: No, you are right, that would work

PR #351, opened 2020-02-07
Fix PanAm tagging

moltude:
Replace open/close quotes with neutral double quotes
Replace literal with URI (@mdellabitta lmk if that is what you had in mind)


moltude:
@mdellabitta 
https://gph.is/18WMfck

mdellabitta:
This PR is "Cleared to Land"!

PR #354, opened 2020-02-21
Add Maine

moltude:
Re-implement mapping from ingesiton1.


src/main/scala/dpla/ingestion3/utils/ProviderRegistry.scala
```diff
@@ -59,6 +59,7 @@ object ProviderRegistry {
     "in" -> Register(profile = new InProfile),
     "lc" -> Register(profile = new LocProfile),
     "ma" -> Register(profile = new MaProfile),
+    "me" -> Register(profile = new MeProfile),
```
AudreyAltman: Aw man, now I'm going to have to remember that the abbreviation for Maine is ME.  If this ever comes up at pub trivia, I will thank you ;)

PR #355, opened 2020-03-03
Stage 1 IIIF

moltude:
All changes implemented stop short of changing what JSON is written for the index. 
* Make `object` multi-valued
* Add dedicated, optional, single-valued URI field for IIIF manifest, `iiifManifest`
* Update row and model converters with these changes
* Update row and model converter tests 
* Update MAP AVRO schema 
* Add comments


moltude:
deprecated branch -- see https://github.com/dpla/ingestion3/pull/356

PR #365, opened 2020-04-26
Feature/wiki identifier enrichment

moltude:
Wikimedia ingest requires Wiki identifiers attached to certain field (e.g. `dataProvider`). Decided that it should probably be an enrichment rather than a function of the Wikimedia export. 

Works essentially like the existing language enrichment and looks up the `EdmAgent.name` value against a list of `value -> wikiId`  values.  Wikimedia identifier is added to the `exactMatch` property of `EdmAgent` even through the value isn't a proper URI. 

The `institutions.json` file was deduped

Add tests and `dataProvider` to enrichment reports/summary. 

🍺 🍺 🍺 🍺


src/main/scala/dpla/ingestion3/enrichments/WikiEntityEnrichment.scala
```diff
@@ -0,0 +1,104 @@
...
+import dpla.ingestion3.utils.FileLoader
+
+/**
+  * Wikimedia entity enrichments
+  */
+class WikiEntityEnrichment extends FileLoader with VocabEnrichment[EdmAgent] {
+
+  protected val wikiUriBase = "https://wikidata.org/wiki/"
+
+  // Files to source vocabulary from
```
mdellabitta, line 13: I wonder if, in the long run, we should be thinking about the lifecycle of this data. How these records get created in Wikidata, and how we put them in the classpath before an ingest. Just a comment to think about, not really about this PR.

src/main/scala/dpla/ingestion3/utils/FileLoader.scala
```diff
@@ -18,6 +22,16 @@ trait FileLoader {
   def getVocabFromCsvFiles(files: Seq[String]): Set[Array[String]] =
     getVocabFromFiles(files).map(_.split(",", 2))
 
+  def getVocabFromJsonFiles(files: Seq[String]): Seq[(String, String)] = {
+    val mapper = new ObjectMapper() with ScalaObjectMapper
+    mapper.registerModule(DefaultScalaModule)
+
+    files.flatMap(file => {
+      val fileContentString = Source.fromInputStream(getClass.getResourceAsStream(file)).getLines().mkString
+      mapper.readValue[Map[String, String]](fileContentString)
```
mdellabitta, line 31: twight.

src/main/scala/dpla/ingestion3/enrichments/WikiEntityEnrichment.scala
moltude: I agree -- this did not seem ideal but good enough for this moment.

PR #379, opened 2020-07-13
Feature/wiki metadata export

moltude:
Fixup nara mapping
- Remove file type filter from mediaMaster mapping
- Fixup typo
Add wiki export 
- Evaluate enriched records for wikimedia eligibility
- Generate wiki markup for eligible records and save to disk inside JSON block
- Uses block list to filter out naids already uploaded to wikimedia


src/test/scala/dpla/ingestion3/wiki/WikiMapperTest.scala
```diff
@@ -1,15 +1,14 @@
...
-import dpla.ingestion3.data.EnrichedRecordFixture
-import dpla.ingestion3.model.{URI, _}
+import dpla.ingestion3.model._
 import org.scalatest.{FlatSpec, Matchers}
 
 /**
   * Tests for the ingestion3 FileIO utility
   */
-class WikiMapperTest extends FlatSpec with Matchers {
+class MapperTest extends WikiMapper
```
mdellabitta, line 9: Curious why the nested classes. Missed this until now. Also, that comment seems to not belong here!

moltude: hmm  -- I thought I moved that nested class out in the last commit. Should be outside the test class now.

PR #382, opened 2020-07-27
Fix/nara edm rights

moltude:
Bring NARA rights mapping in line with https://docs.google.com/spreadsheets/d/1IRsQ9dbCCykItDQE9RnlpWcEMKq0rBjOdqrClSacWSw/edit?usp=drive_web&ouid=110000082651944562863

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/NaraMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/NaraMapping.scala
@@ -462,12 +462,12 @@ class NaraMapping extends XmlMapping with XmlExtractor {
     (parents ++ mediaTypes).distinct
   }
 
-  private def extractRights(data: NodeSeq): Seq[String] = for {
+  private def extractRights(data: NodeSeq): Seq[String] = (for {
     useRestriction <- data \ "useRestriction"
     value1 = (useRestriction \ "note").text
     value2 = (useRestriction \ "specificUseRestrictionArray" \ "specificUseRestriction" \ "termName").text
     value3 = (useRestriction \ "status" \ "termName").text
-  } yield Seq(value2, value1, value3).filter(_.nonEmpty).mkString(" ; ")
+  } yield Seq(value2, value1, value3).filter(_.nonEmpty).mkString(" ; ")).filter(_.nonEmpty)
 
 
   private def extractTypes(data: NodeSeq): Seq[String] = for {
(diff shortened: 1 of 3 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/NaraMapping.scala
```diff
@@ -85,36 +86,40 @@ class NaraMapping extends XmlMapping with XmlExtractor {
...
+        case (Some("Unrestricted"), _) => Some(rightsUri("NoC-US"))
         case (_, _) => None
       }}
 
     // The most restrictive statements
     val mostRestrictive = Seq(
-      Option(URI("http://rightsstatements.org/vocab/InC/1.0/")), // in copyright
-      Option(URI("http://rightsstatements.org/vocab/CNE/1.0/")) // copyright not evaluated
+      Option(rightsUri("InC")), // in copyright
+      Option(rightsUri("CNE"))  // copyright not evaluated
```
DominicBM, line 122: I think this "CNE" should have been switched to "UND" as well?

src/main/scala/dpla/ingestion3/mappers/providers/NaraMapping.scala
```diff
@@ -85,36 +86,40 @@ class NaraMapping extends XmlMapping with XmlExtractor {
...
+        case (Some("Restricted - Possibly"), Some("Trademark")) => Some(rightsUri("NoC-OKLR"))
+        case (Some("Restricted - Possibly"), Some("Other")) => Some(rightsUri("NoC-OKLR"))
+        case (Some("Restricted - Possibly"), _) => Some(rightsUri("UND"))
+        case (Some("Undetermined"), _) => Some(rightsUri("CNE"))
+        case (Some("Unrestricted"), _) => Some(rightsUri("NoC-US"))
         case (_, _) => None
       }}
 
     // The most restrictive statements
     val mostRestrictive = Seq(
```
DominicBM, line 120: @moltude We might need a hierarchy rather than just a single "most restrictive". It occurs to me that you can have any combination of specific use restrictions with three different rights statements—UND, NKC, and NoC-OKLR. If you have UND plus either of the other two (or all three), UND is definitely the correct fallback. Don't we need to resolve the situation where you have NKC + NoC-OKLR (but not UND) as well?

src/main/scala/dpla/ingestion3/mappers/providers/NaraMapping.scala
moltude: Honestly I don't know -- just following the spec here. If you say so I can change it

DominicBM: Yes, I am just suggesting that all CNE/UND should have been  switched in the file, since their meanings were exactly reversed originally. Including here, and also in  the commented section at line 139 & 141. CNE is never going to be part of the "most restrictive" default conversation any more, because it doesn't map to any of the specific use restrictions anymore, only UND does.

PR #406, opened 2021-01-15
Rewrite orbis mapping

moltude:


src/main/scala/dpla/ingestion3/mappers/providers/OrbisCascadeMapping.scala
```diff
@@ -1,127 +1,135 @@
...
   // done
   override def `type`(data: Document[NodeSeq]): Seq[String] =
-    extractStrings(metadataRoot(data) \ "type")
+    extractStrings(data \\ "SourceResource" \ "hasType")
 
   // OreAggregation
   override def dplaUri(data: Document[NodeSeq]): ZeroToOne[URI] = mintDplaItemUri(data)
 
-  // TODO FIXME - TBD could be source or could be publisher
+  // FIXME: Need to dereference the URI
```
mdellabitta, line 85: Don't we blow up if we don't have a dataProvider?

moltude: yes -- this is what I'm asking them to fix and they are opting for a texas style provider lookup table (including text labels is out of scope for remediation).

src/test/scala/dpla/ingestion3/mappers/xml/XmlExtractorUtilsTest.scala
```diff
@@ -0,0 +1,28 @@
...
+                               xmlns:dcterms="http://www.w3.org/1999/xhtml"
+                               xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns"
+                               xmlns:doap="http://usefulinc.com/ns/doap"
+                               xmlns:edm="http://rs.org/">
+      </ore:Aggregation>
+
+    println("metadata " + xml.head.getNamespace("rdf"))
+    println("metadata " + xml.head.getNamespace("edm"))
+
+    val rdfNs = xml \ "@{xmlns}rdf}"
```
mdellabitta, line 21: I don't actually know: Is that trailing `}` after `rdf` correct?

mdellabitta (approved):

PR #421, opened 2021-02-23
Trap all errors during edmRights normalization

moltude:


mdellabitta (commented):
Is the only real change here on line 77?

moltude:
Yeah, I moved the opening of the Try block up to line 44 so the rest of it is all indentation.

mdellabitta (approved):

PR #425, opened 2021-04-08
Fixup wiki markup to comply with CC licenses and permissions

moltude:


mdellabitta (approved):
Just thinking out loud: if the wikitext generation gets more complicated, we might want to use a templating language like https://www.playframework.com/documentation/2.2.x/ScalaTemplates or https://velocity.apache.org/

moltude:
@mdellabitta that would a good tool to throw in the mix

moltude:
I'm going to hold on merging this PR until @DominicBM has a chance to sign off on the revised wiki markup

DominicBM (approved):
Output looks good, based on the documentation! I’m actually hoping the complexity decreased over time. Once we are providing structured data, we need to worry less about formatted text.

PR #427, opened 2021-04-21
NWH name + sbt fixes

moltude:


build.sbt
```diff
@@ -1,66 +1,63 @@
...
-  .settings(
-    name := "ingestion3",
-    version := "0.0.1",
-    scalaVersion := "2.11.8",
-    Defaults.itSettings,
-    parallelExecution in Test := false,
-    libraryDependencies ++= Seq(
-      "org.scalatest" %% "scalatest" % "3.0.1" % "it,test",
+libraryDependencies ++= Seq(
+  // "org.scalatest" %% "scalatest" % "3.0.1" % "it,test",
```
mdellabitta, line 15: where is scalatest coming from? we should ban it from there and pin our version rather than accepting whatever a different project sticks us with

mdellabitta (changes requested):

build.sbt
moltude: i should have just removed `it` because the `it` package was removed from settings (not used afaik)

mdellabitta: i think mark had something in there to test the geocoding stuff at one point, but...

mdellabitta (approved):

PR #428, opened 2021-05-05
Ohio + NJDE

moltude:
Initial mapping for NJ/DE hub
Ohio file harvest


src/main/scala/dpla/ingestion3/harvesters/file/OaiFileHarvester.scala
```diff
@@ -153,6 +153,8 @@ class OaiFileHarvester(spark: SparkSession,
       IOUtils.closeQuietly(inputStream)
     })
 
+    flush()
```
mdellabitta, line 156: :toilet:

mdellabitta (approved):

PR #429, opened 2021-05-18
Trash panda party

moltude:
Hot mess of a PR
- Add initial mapping for NJ/DE
- Mapping fix for NWDH that filters lat/long coordinates from spatial mapping [this should probably be the new "spatial enrichment"]
- Mapping fix for NWDH to map intermediate provider
- Explainer for UW exclusion
- SBT spark packages fix
- Ohio Hub now uses OAI file harvester


mdellabitta (approved):
:+1:

PR #454, opened 2021-11-15
Feature/nypl mapper

moltude:


mdellabitta (approved):
Little curious what this null default value for the constructor does. Also seems like you could get the XML out of the field with just calling toString() on the JValue you get out of it rather than making another JsonExtractor.

That seems like cleanup stuff (if I'm right about it, which ¯\_(ツ)_/¯) and this seems like it will work well either way!

moltude:
Ah -- that default value was cruft from earlier iterations of trying to reused existing trails in some other variations. It is a no op with current implementation. I'll remove

Yeah, I don't know about that toString() call, I might not be doing exactly what you have in mind but it didn't work for me.

PR #484, opened 2022-05-12
Add Jewish Heritage Network

moltude:
Initial mapping and harvester

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/utils/Extractor.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/utils/Extractor.scala
@@ -2,7 +2,8 @@ package dpla.ingestion3.mappers.utils
 
 import org.json4s.JsonAST._
 
-import scala.xml.{Elem, Node, NodeSeq, Text}
+import scala.util.{Try, Success, Failure}
+import scala.xml.{Elem, MetaData, Node, NodeSeq, Text}
 
 
 /**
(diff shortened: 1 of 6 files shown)


src/main/scala/dpla/ingestion3/mappers/utils/Extractor.scala
```diff
@@ -134,6 +147,24 @@ trait XmlExtractor extends Extractor[NodeSeq] {
...
     node.child.collect{ case Text(t) => t }.map(_.trim).filterNot(_.isEmpty)
   }
+
+  /**
+   *
+   * @param node
+   * @param attribute
+   * @return
+   */
+  def getAttributeValue(node: Node, attribute: String): Option[String] = {
```
mdellabitta, line 157: do you need this? Doesn't `\@` work?

moltude: No, it doe not with with name spaced attributes for which the namespace is not declared

mdellabitta: Should we force some well-known prefixes into the document in the Mapper so this doesn't happen?

moltude: Ohh that is a great idea.

moltude: Added list of predefined namespaces to extractor. It won't help in this case but it will for other providers where namespaces are used for attributes and the namespaces are declared in the document.

PR #488, opened 2022-06-14
Add mapping for MDL iiif manifest

moltude:
Add IIIF mapping for MDL


moltude:
@DominicBM Do we need to confirm with MDL about the base URL path for IIIF manfiests? 

`protected lazy val iiifManifestBase = "https://cdm16022.contentdm.oclc.org"`

mdellabitta (approved):

PR #522, opened 2022-10-12
DLG spatial mapping

moltude:
- Update place/spatial mapping for DLG to add coordinates and URIs to the last place in the names array
- Update DPLA MAP model with extactMatch property
- Fixup (deprecated) spatial enrichment

### DIFF
--- a/src/main/scala/dpla/ingestion3/model/DplaMapData.scala
+++ b/src/main/scala/dpla/ingestion3/model/DplaMapData.scala
@@ -128,7 +128,8 @@ case class DplaPlace(
                       state: ZeroToOne[String] = None,
                       country: ZeroToOne[String] = None,
                       region: ZeroToOne[String] = None,
-                      coordinates: ZeroToOne[String] = None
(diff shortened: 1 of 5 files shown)


src/main/scala/dpla/ingestion3/model/DplaMapData.scala
```diff
@@ -128,7 +128,8 @@ case class DplaPlace(
                       state: ZeroToOne[String] = None,
                       country: ZeroToOne[String] = None,
                       region: ZeroToOne[String] = None,
-                      coordinates: ZeroToOne[String] = None
+                      coordinates: ZeroToOne[String] = None,
+                      exactMatch: ZeroToMany[URI] = Seq()
```
mdellabitta, line 132: GHOST FIELD 👻

mdellabitta (approved):

moltude:
IDK how I mucked it so bad, I swear I ran the tests and they passed before I pushed this..

PR #523, opened 2022-10-17
Mapping/Model + Row converters for DplaPlace.exactMatch and DcmiCollection.isShownAt 

moltude:
* Add DcmiCollection.isShownAt to DplaMapModel, row and model converter
* Add DplaPlace.exactMatch to model and row converters 
* Add tests for Model and Row converters for DplaPlace and DcmiCollection 
* Add mapping to NARA for DcmiCollection.isShownAt
* Fix up base URI for NARA URIs (catalog.archives.org -> catalog.archives.gov)

### DIFF
--- a/src/main/scala/dpla/ingestion3/mappers/providers/NaraMapping.scala
+++ b/src/main/scala/dpla/ingestion3/mappers/providers/NaraMapping.scala
@@ -15,7 +15,7 @@ import scala.xml.{Node, NodeSeq}
 
 class NaraMapping extends XmlMapping with XmlExtractor {
 
-  private val uriBase = "https://catalog.archives.org/id/"
+  private val uriBase = "https://catalog.archives.gov/id/"
 
   private def uri(naid: String): URI = URI(s"$uriBase$naid")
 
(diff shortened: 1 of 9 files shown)


src/main/scala/dpla/ingestion3/mappers/providers/NaraMapping.scala
```diff
@@ -247,22 +247,34 @@ class NaraMapping extends XmlMapping with XmlExtractor {
     uri = Some(URI("http://dp.la/api/contributor/nara"))
   )
 
-  private def extractCollection(data: NodeSeq): Seq[String] = {
+  private def extractCollection(data: NodeSeq): Seq[DcmiTypeCollection] = {
     val parentRecordGroupIds = for {
       prg <- data \\ "parentRecordGroup" \\ "title"
-    } yield prg.text
+      prgUri <- extractString(data \\ "parentRecordGroup" \\ "naId")
```
mdellabitta, line 253: Not the end of the world, but it's probably more idiomatic to use `=` in a `for` rather than chaining `map`.

src/main/scala/dpla/ingestion3/model/RowConverter.scala
```diff
@@ -1,6 +1,7 @@
 package dpla.ingestion3.model
 
 import dpla.ingestion3.messages.IngestMessage
+import dpla.ingestion3.model.RowConverter.fromEdmWebResource
```
mdellabitta, line 4: I think you don't need this?

mdellabitta (approved):

mdellabitta:
Left a couple of comments that aren't that important!

PR #531, opened 2022-11-06
Updates for changes rights

DominicBM:
This patch should make the following 3 changes:

1. In the case of a NoC-US item, the `permission` template parameter now uses `NoC-US | $dataProviderWikiUri` instead of `PD-US`.
2. New case of a `.../NKC/` item is now allowed, for which the `permission` template parameter uses `NKC | $dataProviderWikiUri`.
3. Combining the rights statement with the permission field means we can simplify the output by removing the `Other fields` portion of the template, and the code required to check `getWikiOtherFieldsRights` code required to produce it.

I have updated the test case so it reflects a valid current output based on above changes.

### DIFF
--- a/src/main/scala/dpla/ingestion3/model/package.scala
+++ b/src/main/scala/dpla/ingestion3/model/package.scala
@@ -250,17 +250,6 @@ package object model {
         | }}""".stripMargin
   }
 
-  def getWikiOtherFieldsRights(edmRights: Option[URI]): String =
-    edmRights match {
-      case Some(uri) =>
-        if(uri.toString.startsWith("http://rightsstatements.org"))
-          s"{{ InFi | Standardized rights statement | {{ rights statement | ${escapeWikiChars(uri.toString)} }} }}"
-        else
-          ""
-      case None => ""
-    }
-
-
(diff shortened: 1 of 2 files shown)


src/main/scala/dpla/ingestion3/model/package.scala
```diff
@@ -247,25 +247,14 @@ package object model {
...
-          ""
-      case None => ""
-    }
-
-
   def getWikiPermissionTemplate(edmRights: Option[URI]): String = {
     edmRights match {
       case Some(uri) => uri.toString match {
-        case t if t.startsWith("http://rightsstatements.org/vocab/NoC-US/") => "PD-US"
+        case t if t.startsWith("http://rightsstatements.org/vocab/NKC/") => "NKC | $dataProviderWikiUri"
```
moltude, line 256: @DominicBM this shouldn't be adding the $dataProviderWiki URI here because it doesn't have access to that value in the method.

moltude: It should be added in the markup function `buildWikiMarkup`

moltude (changes requested):

moltude (approved):

PR #532, opened 2022-11-07
Nara memory efficiency

AudreyAltman:
Last week we discussed making the temp file locations configurable.  Does that still seem like the right move?  And if so, what's the best way to do this - input variables?


moltude (approved):

mdellabitta:
So I think EMR spins up HDFS by default, and the path you hardcoded would resolve to HDFS if a schema isn't otherwise specified in the file URL. I'd say the only other thing to think about is perhaps concatenating a job id or something random in the path so multiple runs on the same cluster or box don't collide with one another.

PR #559, opened 2023-10-18
Add wiki identifier for provider values in the wiki markup

moltude:


src/main/scala/dpla/ingestion3/model/package.scala
```diff
@@ -248,7 +249,7 @@ package object model {
         |   | permission = {{${permissions}}}
         |   | source = {{ DPLA
         |       | ${escapeWikiChars(dataProviderWikiUri)}
-        |       | hub = ${escapeWikiChars(record.provider.name.getOrElse(""))}
+        |       | ${escapeWikiChars(providerWikiUri)}
```
DominicBM, line 252: Should still have `hub = ` here

DominicBM (changes requested):
I couldn't test the actual function that was added, but assuming it outputs the WD ID as needed, the only change is at the line I indicated, where the output should still include `hub = ` in the wiki markup, with the ID ont he same like after that.

src/main/scala/dpla/ingestion3/model/package.scala
moltude: Got it, thanks. I assumed it should have looked like the dataProviderWikiId w/o the `hub = ` bit. Will revise.

DominicBM (approved):

PR #571, opened 2024-02-22
Bugfix for "Harvest all records" failure in dpla-combine

cedelis:
In order to read local /tmp file instead of hdfs file. Otherwise, I get the following error:

'Path does not exist: hdfs://hadoop-namenode/tmp/oai7424538003476557972.txt;'


mdellabitta (approved):

mdellabitta:
I had to cherry pick this change because of formatting changes on develop that were making merges awkward: https://github.com/dpla/ingestion3/commit/1e4a1eebbc4ada4370c632b1d9f48bf5dd4f9696

Thanks for contributing this!

PR #780, opened 2026-09-15
Issue #776 Fix PrimoVEHarvester offset: switch to 0-based paging

### LINKED ISSUE #776: PrimoVEHarvester skips the first record of every result set (offset is 1-based, API is 0-based)
## Summary

`PrimoVEHarvester.harvest` starts paging at `offset = "1"`, but the Primo VE search API's `offset` is **0-based**. The first record of every result set is silently skipped.

**This is not currently causing data loss in production, and there is nothing to do right now.** All three Primo hubs are on other paths (see "Why this isn't urgent"). Filing it so the bug is on record if the Scala Primo harvesters are ever brought back into use.

## Evidence

Measured against Getty's live Primo VE API (from the allowlisted IP), `limit=3`:

```
offset=0  →  GETTY_ROSETTAIE10176553, GETTY_ROSETTAIE102335,  GETTY_ROSETTAIE103576
offset=1  →  GETTY_ROSETTAIE102335,   GETTY_ROSETTAIE103576,  GETTY_ROSETTAIE10309750
```

`…IE10176553` is never retrieved. Both offsets return HTTP 200, so nothing surfaces in logs.

## Affected

- `GettyHarvester` → `PrimoVEHarvester`
- `MississippiHarvester` → `PrimoVEHarvester`

**Not** `MwdlHarvester` — it extends `PrimoHarvester` (Primo *Classic*), which uses `indx`, and `indx` is legitimately 1-based there. No change needed in the Classic path.

## The fix is not a one-liner

`src/main/scala/dpla/ingestion3/harvesters/api/PrimoVEHarvester.scala`. Three things assume 1-based and have to move together:

1. `var offset = "1"` (line ~51) → `"0"`, and the `"offset" -> Some("1")` default in `queryParams` (line ~36).
2. The progress log computes `nextIndx.toLong - 1` (line ~91) to convert 1-based to a count — that `- 1` becomes wrong once offset is 0-based.
3. The termination test `if (offset.toInt >= totalRecords.toInt)` (line ~96) is off by one in the 0-based world; worth re-checking it doesn't drop the final page.

Worth a test that asserts the first record of a known result set is present.

## Why this isn't urgent

All three Primo hubs are currently on other paths, so this code is effectively dormant:

- **Mississippi** — harvested via out-of-band Python prefix scripts (`mississippi-harvest.py`, unmerged branch `feat/mississippi-prefix-harvest`) → JSONL → `jsonl-to-avro.py`. Those scripts start at `offset = 0` and are correct.
- **MWDL** — also on out-of-band Python; the Scala path targets the retired Primo Classic host (see below).
- **Getty** — the Scala path *is* what `ingest.sh getty` runs, but it is blocked upstream by a hard Ex Libris offset cap at 2,000 records (cold request at `offset=2000` → `403 ROUTING_ERROR`), so it can only ever retrieve ~2,000 of ~101,393. `getty.schedule.status` is `on-hold`.

## Related, same "dormant old harvester" condition

Worth looking at together if anyone revives this code:

- **`MwdlHarvester` targets a retired API generation.** It builds Primo *Classic* URLs against a hardcoded `utah-primoprod.hosted.exlibrisgroup.com/PrimoWebServices/xservice/search/brief`. MWDL migrated to Primo VE on 2026-08-06, and there is no `mwdl.harvest.endpoint` key in `i3.conf` to repoint it. `ingest.sh mwdl` is dead against that host.
- **Page-size config is dead in both VE harvesters.** `PrimoVEHarvester` puts `conf.harvest.rows` into `queryParams` under the key `"rows"`, but neither subclass reads it: `GettyHarvester` hardcodes `limit=500`, and `MississippiHarvester` reads `params.getOrElse("limit", "100")` — a `"limit"` key that is never populated. So `getty.harvest.rows="100"` and `mississippi.harvest.rows` have no effect; the real page sizes are 500 and 100 respectively.
- **Retry/backoff is too aggressive for this API.** `HttpUtils` retries 5× with 1s/2s/4s/8s backoff (~15s total). The Python harvesters that actually work against Primo wait 12–20s *between every successful request* and 300s on a throttle.
- **A mid-harvest error yields a partial harvest that exits 0.** The loop logs the error, sets `continueHarvest = false`, and returns whatever was written. Downstream this looks like a successful harvest; only the anomaly/>5% drop gate catches it, so a *small* partial failure could pass unnoticed.

zspieges:
- queryParams default and loop init: "1" → "0"
- Progress log: remove spurious - 1 (count was off by one)
- Termination: check nextIndx >= totalRecords instead of offset, avoiding an extra empty API call on the last page

## Summary

Primo VE harvesting now uses zero-based paging. `PrimoVEHarvester`, `GettyHarvester`, and `MississippiHarvester` default to offset `0`. `PrimoVEHarvester` stops when the next index reaches or exceeds `totalRecords`, which prevents an extra empty API call.

## Operational impact

- No manual pipeline trigger or ECS redeployment is indicated.
- No environment variables, AWS Secrets Manager keys, database migrations, shared infrastructure changes, public API changes, or security-sensitive changes are indicated.


zspieges:
@coderabbitai review

zspieges:
@coderabbitai full review

